PR #9, opened 2020-05-17
RPRBLND-1338: Back plates does not work correctly 

bnagirniak:
PURPOSE
1. When disabling background override even if the file is still loaded, the backplate should not show up in production render. 
2. Backplates should work in viewport render.
3. Code related to backplate could be simplified and improved.

EFFECT OF CHANGE
1. Fixed disabling backplate when background override is disabled.
2. Made backplate work in Viewport render.
3. Added "Crop" setting for Backplate panel, which provides option to user to crop background image due to render resolution.

TECHNICAL STEPS
1. Code refactoring:
- simplified image.ImagePixels class, made it to export image including tile.
- moved Backplate class to WorldData internal class as BackplateData. Included its data into comparison of WorldData. Also simplified code in BackplateData and adjusted code world export design.
2. Added backplate_crop to world properties.
3. Enabled backplate in viewport render.

### DIFF
--- a/src/rprblender/export/image.py
+++ b/src/rprblender/export/image.py
@@ -129,45 +129,34 @@ class ImagePixels:
 
         return region_pixels
 
-    def export_full(self, rpr_context) -> (pyrpr.Image, None):
-        """ Export the full image pixels as RPR image"""
-        if self.is_empty():
-            return None
-
-        image_key = f"{self.name}@{self.color_space}"
-
-        if image_key in rpr_context.images:
-            return rpr_context.images[image_key]
-
-        pixels = np.ascontiguousarray(np.flipud(self.pixels))
-        rpr_image = rpr_context.create_image_data(image_key, pixels)
-        rpr_image.set_name(image_key)
-
-        if self.color_space in ('sRGB', 'BD16', 'Filmic Log'):
-            rpr_image.set_gamma(2.2)
-
-        return rpr_image
-
-    def export_region(self, rpr_context, x1, y1, x2, y2) -> (pyrpr.Image, None):
+    def export(self, rpr_context, region=None) -> (pyrpr.Image, None):
         """ Export pixels cropped to sub-region coordinates as RPR image """
         if self.is_empty():
             return None
 
-        # check sub-region boundaries, just in case something went terribly wrong
-        if x1 == x2 or y1 == y2 or x1 >= self.size[0] or x2 < 0 or y1 >= self.size[1] or y2 < 0:
-            log.warn(f"Image region ({x1}; {y1})-({x2}; {y2}) has no data", self.name)
-            return None
+        if region:
+            x1, y1, x2, y2 = region
+            # check sub-region boundaries, just in case something went terribly wrong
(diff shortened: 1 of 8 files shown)


src/rprblender/export/image.py
```diff
@@ -92,84 +90,53 @@ def sync(rpr_context, image: bpy.types.Image):
...
+            x2, y2 = image_size
 
-        rpr_image = rpr_context.create_image_data(image_key, np.ascontiguousarray(pixels))
+        x1, y1, x2, y2 = (x1 + (x2 - x1) * tile[0][0],
+                          y1 + (y2 - y1) * tile[0][1],
+                          x1 + (x2 - x1) * (tile[0][0] + tile[1][0]),
+                          y1 + (y2 - y1) * (tile[0][1] + tile[1][1]))
 
-        rpr_image.set_name(image_key)
+        pixels = self.pixels[int(y1): int(y2), int(x1):int(x2), :]
```
ardraeiss, line 137: If image or tile size is small enough the region coordinates would end with 1-pixel width or height, resulting in empty pixels. It will cause the CoreError in the next line.
The [int(y1):int(y2)+1, int(x1):int(x2)+1, :] would fix that.

ardraeiss: test case: 3*3 image, 800*900 render size, tiles

bnagirniak: I think adding +1 isn't so correct, that's why added check if y1=y2 or x1=x2 and return None in this case.

src/rprblender/export/image.py
```diff
@@ -92,84 +90,58 @@ def sync(rpr_context, image: bpy.types.Image):
...
-        return rpr_image
+        else:
+            x1, y1 = 0, 0
+            x2, y2 = image_size
 
-    def export_region(self, rpr_context, x1, y1, x2, y2) -> (pyrpr.Image, None):
-        """ Export pixels cropped to sub-region coordinates as RPR image """
-        if self.is_empty():
-            return None
+        x1, y1, x2, y2 = (
```
bsavery, line 132: Just a comment, not really sure what assigning these in a tuple fashion gains, but looks fine ;)

src/rprblender/properties/world.py
```diff
@@ -238,6 +238,11 @@ class RPR_EnvironmentProperties(RPR_Properties):
         description="Background override type",
         default='SPHERE',
     )
+    backplate_crop: bpy.props.BoolProperty(
```
bsavery, line 241: another semantic point, but maybe should change the description of this property. This is actually a resize, not a crop, no?

src/rprblender/export/image.py
bnagirniak: I'm assigning here image rectangle coordinates due to tile values.

src/rprblender/properties/world.py
bnagirniak: It is more related to cropping edges of backplate image for keeping aspect ratio of the image in render. That's why I think crop better corresponds here instead resize.

PR #21, opened 2020-06-05
Add support for MULTIPLY_ADD mode of math Cycles node

cmdrf:
There are a few more modes with three operands that are still missing:
- `SMOOTH_MIN`
- `SMOOTH_MAX`
- `WRAP`
- `COMPARE`

### DIFF
--- a/src/rprblender/nodes/blender_nodes.py
+++ b/src/rprblender/nodes/blender_nodes.py
@@ -1148,7 +1148,11 @@ class ShaderNodeMath(NodeParser):
             elif op == 'MODULO':
                 res = self.create_arithmetic(pyrpr.MATERIAL_NODE_OP_MOD, in1, in2)
             else:
-                raise ValueError("Incorrect math operation", op)
+                in3 = self.get_input_value(2)
+                if op == 'MULTIPLY_ADD':
+                    res = in1 * in2 + in3
+                else:
+                    raise ValueError("Incorrect math operation", op)
 
         if self.node.use_clamp:
             res = res.clamp()


src/rprblender/nodes/blender_nodes.py
```diff
@@ -1148,7 +1148,11 @@ def export(self):
             elif op == 'MODULO':
                 res = self.create_arithmetic(pyrpr.MATERIAL_NODE_OP_MOD, in1, in2)
             else:
-                raise ValueError("Incorrect math operation", op)
+                in3 = self.get_input_value(2)
```
bsavery, line 1151: Are there other 3 input math operations?

cmdrf: Yes, see the description. `WRAP` and `COMPARE` are probably easy to do. But I have no idea what `SMOOTH_MIN` and `SMOOTH_MAX` do exactly.

bsavery: https://www.iquilezles.org/www/articles/smin/smin.htm There's a bunch of diffferent ways to calculate it.  

It looks like blender does:
`ccl_device_inline float smoothminf(float a, float b, float k)
{
  if (k != 0.0f) {
    float h = fmaxf(k - fabsf(a - b), 0.0f) / k;
    return fminf(a, b) - h * h * h * k * (1.0f / 6.0f);
  }
  else {
    return fminf(a, b);
  }
}`

src/rprblender/nodes/blender_nodes.py
```diff
@@ -1148,7 +1148,11 @@ def export(self):
...
+                if op == 'MULTIPLY_ADD':
+                    res = in1 * in2 + in3
+                else:
+                    raise ValueError("Incorrect math operation", op)
```
ardraeiss, line 1155: With so many new Math and Vector Math options it would be much better to use something like
                log.warn(f"Ignoring unsupported Math operation {op}")
                return None
instead of raising ValueError

bnagirniak: Agree with such proposition

bsavery:
Hi @cmdrf I made a few changes.  1st of all the PR should be done against the "develop" branch instead of master.  no worries there.  Also I'll be adding more math ops in the future.

bsavery:
Thanks for your contribution!

PR #23, opened 2020-06-11
Reverting back using image cache instead loading images from memory

bnagirniak:
PURPOSE
Loading images from blender memory could significantly slowdown export time for some scenes, even using fast foreach_get() method to get image data.

EFFECT OF CHANGE
Reverted back using image cache instead loading images from memory (relates to Blender 2.83+), which should increase speed of export for some scenes.

TECHNICAL STEPS
Made caching images to go before loading from blender memory.

### DIFF
--- a/src/rprblender/export/image.py
+++ b/src/rprblender/export/image.py
@@ -64,19 +64,13 @@ def sync(rpr_context, image: bpy.types.Image, use_color_space=None):
     log("sync", image)
 
     pixels = image.pixels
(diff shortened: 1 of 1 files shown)


bsavery:
As I mentioned on slack, why not just disable the foreach_get method for export engine?  In other engines this should be faster no?

bnagirniak:
Oh, seems I didn't understand you correctly. Sure, I can just disable foreach_get only for export engine.

PR #28, opened 2020-06-24
RPRBLND-1381 Mute nodes

bsavery:
PURPOSE
Support "Mute Node" key (M)

EFFECT OF CHANGE
Muting nodes has desired effect:
A muted node passes through the first matching input (i.e if output is color, get the first color input).  Or Returns black if that input is not linked.  If No matching input, use the incoming socket default value.


src/rprblender/nodes/node_parser.py
```diff
@@ -151,6 +151,22 @@ def get_input_link(self, socket_key: [str, int], accepted_type=None):
...
 
+            # check if input node muted
+            if link.from_node.mute:
+                # pass through first instance of same socket type
+                matching_incoming_socket = next((input for input in link.from_node.inputs
+                            if isinstance(input, type(link.from_socket))), None)
+                if matching_incoming_socket:
+                    if matching_incoming_socket.is_linked:
+                        # pass through first incoming link if there is one
+                        link = matching_incoming_socket.links[0]
```
ardraeiss, line 162: Linked node can be muted as well.

bsavery:
> I propose to implement this task in following way:
> 
> 1. Add function export_muted() which should find appropriate input from another linked node.
> 2. In final_export() add check:
> 
> ```
> if self.node.mute:
>     node_item = self.export_muted()
> elif if isinstance(self.rpr_context, RPRContextHybrid):
> ... 
> ```
> 
> Such way is more easier and will fix problem with muted linked node.

I see what you're saying there.  However I see one problem.  That it would need to be implemented for BaseNodeParser class AND NodeParser class.  Not a huge problem.

src/rprblender/nodes/node_parser.py
```diff
@@ -273,7 +293,9 @@ def final_export(self):
         """
 
         log("export", self.material, self.node, self.socket_out, self.group_nodes)
-        if isinstance(self.rpr_context, RPRContextHybrid):
+        if self.node.mute:
+            return self.export_muted()
```
ardraeiss, line 297: Fails with
...
  File "D:\Work\RPR\blender_280\RadeonProRenderBlenderAddon\src\bindings\pyrpr\.build\pyrprwrap.py", line 1393, in ShapeSetMaterial
    return lib.rprShapeSetMaterial(shape._get_handle() if shape else ffi.NULL, node._get_handle() if node else ffi.NULL)
AttributeError: 'NodeItem' object has no attribute '_get_handle'
if entry shader node is muted, for example Mix Shader node connected to Output.

src/rprblender/nodes/node_parser.py
```diff
@@ -244,13 +244,33 @@ def export(self):
...
+        # pass through first instance of same socket type
+        matching_incoming_socket = next((input for input in self.node.inputs
+                    if isinstance(input, type(self.socket_out))), None)
+        if matching_incoming_socket:
+            if matching_incoming_socket.is_linked:
+                # pass through first incoming link if there is one
+                return self.get_input_link(matching_incoming_socket.name)
+            else:
+                # return black
+                return NodeItem(self.rpr_context, (0.0, 0.0, 0.0, 0.0))
```
bnagirniak, line 258: BaseNodeParser doesn't work with NodeItem, therefore here we should return just (0.0, 0.0, 0.0, 0.0).
Also export_muted() function should be overrided in NodeItem in the same way as get_input_link().

src/rprblender/nodes/node_parser.py
bnagirniak, line 297: Use node_item = self.export_muted() here, because it is going to be processed further in this function.

src/rprblender/nodes/node_parser.py
```diff
@@ -246,16 +246,11 @@ def export(self):
...
-                    if isinstance(input, type(self.socket_out))), None)
+            if isinstance(input, type(self.socket_out)) and input.is_linked), None)
         if matching_incoming_socket:
-            if matching_incoming_socket.is_linked:
-                # pass through first incoming link if there is one
-                return self.get_input_link(matching_incoming_socket.name)
-            else:
-                # return black
-                return (0.0, 0.0, 0.0, 0.0)
+            return self.get_input_link(matching_incoming_socket.name)
```
bsavery, line 253: So my testing showed that in cycles if there was a matching socket but it wasn't linked it would return black. 

In the above code I think it would us the input value?  Correct me if I'm wrong.

bnagirniak: Hmm, my tests in cycles showed me that if nothing is connected to matching socket in muted node, than it works like nothing is connected (like muted node doesn't exists) and it uses default value.

bsavery: It's a bit of a weird use case anyway.  Ok.

PR #30, opened 2020-06-25
Rprblnd 1380. Add support for Local mode "/" key in viewport

bsavery:
PROBLEM:  Blender has a "Local mode" to viewport which highlights the selected object only.  RPR does not respect this.

EFFECT:  User sees only selected object when pressing / key.

NOTE:  Uses the function visible_in_viewport_get() mainly


aledbrown:
Thank you for adding this one, it will be so handy to have local mode working in iPR. It is on my list to add it here and I really meant to come sort it before now. I have a couple more issues to report as well.

PR #35, opened 2020-07-10
Rprblnd-1382. Mapping node and generated lookup

bsavery:
PURPOSE:  Support UV mapping transformations better

EFFECT:  
UV lookup "Generated" works better
Vector mapping math is more correct in most instances.  In particular before it was scaling/rotating in wrong order and rotating by x or y did not work

Review notes:
Generated lookup is simply P_Local mapped to 0-1 hence we get that from normalizing by bounding box

Mapping node follows method from https://github.com/blender/blender/blob/master/intern/cycles/kernel/shaders/node_mapping.osl


src/rprblender/nodes/blender_nodes.py
```diff
@@ -972,42 +973,54 @@ def export_hybrid(self):
...
+        if tex_coord_type == 'Generated':
+            data = self.create_node(pyrpr.MATERIAL_NODE_INPUT_LOOKUP, {
+                pyrpr.MATERIAL_INPUT_VALUE: pyrpr.MATERIAL_NODE_LOOKUP_P_LOCAL,
+            })
+            if self.object:
+                # normalize over object bounding box
+                # get min and max of boinding box (is there a faster way?)
+                min = [sys.maxsize] * 3
+                max = [-sys.maxsize] * 3
+                for point in self.object.bound_box:
```
bsavery, line 992: Is there a faster way to do this in python?

bnagirniak: yes:
mn = tuple(min(p[i] for p in self.object.bound_box) for i in range(3))
mx = tuple(max(p[i] for p in self.object.bound_box) for i in range(3))

src/rprblender/nodes/blender_nodes.py
```diff
@@ -1148,10 +1161,6 @@ def export(self):
             elif op == 'MODULO':
                 res = self.create_arithmetic(pyrpr.MATERIAL_NODE_OP_MOD, in1, in2)
             else:
-                in3 = self.get_input_value(2)
-                if op == 'MULTIPLY_ADD':
-                    res = in1 * in2 + in3
-                else:
```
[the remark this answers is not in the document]
bsavery: mistake.  Will fix.

src/rprblender/nodes/blender_nodes.py
bsavery: thanks.

src/rprblender/nodes/blender_nodes.py
```diff
@@ -972,42 +973,48 @@ def export_hybrid(self):
...
+        tex_coord_type = self.socket_out.name
+
+        if tex_coord_type == 'Generated':
+            data = self.create_node(pyrpr.MATERIAL_NODE_INPUT_LOOKUP, {
+                pyrpr.MATERIAL_INPUT_VALUE: pyrpr.MATERIAL_NODE_LOOKUP_P_LOCAL,
+            })
+            if self.object:
+                # normalize over object bounding box
+                # get min and max of boinding box
+                min = tuple(min(p[i] for p in self.object.bound_box) for i in range(3))
```
bnagirniak, line 990: Have following error here:
```
  File "D:\amd-gpuopen\RadeonProRenderBlenderAddon\src\rprblender\nodes\blender_nodes.py", line 990, in export
    min = tuple(min(p[i] for p in self.object.bound_box) for i in range(3))
  File "D:\amd-gpuopen\RadeonProRenderBlenderAddon\src\rprblender\nodes\blender_nodes.py", line 990, in <genexpr>
    min = tuple(min(p[i] for p in self.object.bound_box) for i in range(3))
NameError: free variable 'min' referenced before assignment in enclosing scope
```
Better not to override built-in min and max functions, suggest to use names mn, mx.

bsavery: fixed

bsavery:
> Z-axis texture rotation result is incorrect in Point mode; correct in Texture.
> 
> Result of any two/three axis rotations simultaneously is always wrong.

I fixed z rotation.  however I don't have a good explanation why 2 axis rotation is different.  This seems closer now, if you have any suggestions to improve, please suggest.

PR #36, opened 2020-07-17
Add support for Object Info node

bsavery:
EFFECT:
Object Info Node works and is added to menu.

Notes:
Uses node.object and node.material extensively.  Only part that doesn't work is random input in hybrid.


src/rprblender/nodes/blender_nodes.py
```diff
@@ -972,6 +972,39 @@ def export_hybrid(self):
         return shader2
 
 
+class ShaderNodeObjectInfo(NodeParser):
+    
+    def export(self):
+        if self.socket_out.name == 'Location':
+            if self.object:
+                return self.node_item(self.object.location)
```
bnagirniak, line 980: should be: return self.node_item(tuple(self.object.location))

src/rprblender/nodes/blender_nodes.py
```diff
@@ -972,6 +972,39 @@ def export_hybrid(self):
...
+            else: 
+                return self.node_item((0.0, 0.0, 0.0, 0.0))
+        elif self.socket_out.name == 'Color':
+            if self.object:
+                return self.node_item(self.object.color)
```
bnagirniak, line 985: should be: return self.node_item(tuple(self.object.color))

src/rprblender/nodes/blender_nodes.py
bsavery: Clearly I didn't test all the inputs.  However I could claim that maybe node_item should take any iterable with the right length.  What do you think?

bsavery: Not just tuple.

bsavery:
> It's fine for single object.
> 
> For multiple objects with the same material it will be a shared constant material instead of "per-object". For example, if "Position" mode is used the first exported object will define the color of all other material users.
> 
> To fix it materials separation per object would be required, like with UV Map node.

How do you propose to fix this?  Maybe This:
https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/blob/0a23e865e81a19bfa818d89aa57380eaacaf8d9f/src/rprblender/export/material.py#L26

Should be generalized to use object as key?

ardraeiss:
> How do you propose to fix this? Maybe This:
> 
> https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/blob/0a23e865e81a19bfa818d89aa57380eaacaf8d9f/src/rprblender/export/material.py#L26
> 
> Should be generalized to use object as key?

Yeah, I think this is the way - to generalize it to something like "unique_material_id_needed(material)".

bsavery:
@ardraeiss  Im going to say your point is valid but can be addressed in another PR to do per object material.

bnagirniak:
> Im going to say your point is valid but can be addressed in another PR to do per object material.

I agree, propose to implement such materials per object (only if they have nodes dependent by object) in separate task. I have ideas how do this quite easy.

PR #38, opened 2020-07-23
Adapt viewport resolution to samples per second

bsavery:
The idea here is to set a "samples per second" for viewport rendering and adapt the resolution of viewport to reach it.  

TODOS:

- I don't quite get all the resolution setting code, and get a crash there.
- We should probably set a hard floor for downresing resolution
- Encorporate viewport denoising?

### DIFF
--- a/src/rprblender/engine/viewport_engine.py
+++ b/src/rprblender/engine/viewport_engine.py
@@ -118,27 +118,29 @@ class ViewportSettings:
             ((self.border[0][0] / self.screen_width, self.border[0][1] / self.screen_height),
              (self.border[1][0] / self.screen_width, self.border[1][1] / self.screen_height)))
 
-    def adapt_resolution(self, requested_pixels, prev_settings, min_scale):
+    def adapt_resolution(self, requested_ratio, prev_settings, min_scale):
         MIN_PIXEL_DIFF = 0.2
         MIN_RATIO_DIFF = 0.1
 
         # checking with previous settings, if previous resolution
         # suits to requested_pixels and suits to new render ratio
-        prev_pixels = prev_settings.width * prev_settings.height
-        if abs(1 - requested_pixels / prev_pixels) < MIN_PIXEL_DIFF \
+        if abs(1 - requested_ratio) < MIN_PIXEL_DIFF \
                 and abs(self.width / self.height -
                         prev_settings.width / prev_settings.height) < MIN_RATIO_DIFF:
-            self.width, self.height = prev_settings.width, prev_settings.height
+            self.width = max(prev_settings.width, int(self.width * min_scale), 1)
+            self.height = max(prev_settings.height, int(self.height * min_scale), 1)
             return
 
         # checking if current resolution suits to requested_pixels
-        pixels = self.width * self.height
-        if requested_pixels > pixels or abs(1 - requested_pixels / pixels) < MIN_PIXEL_DIFF:
+        new_ratio = (prev_settings.width * prev_settings.height * requested_ratio) / \
+                    (self.width * self.height)
+        if new_ratio > 1.0 - MIN_PIXEL_DIFF:
             return
 
         # changing to new resolution, but not less then SCALE_MIN
-        scale = max(math.sqrt(requested_pixels / pixels), min_scale)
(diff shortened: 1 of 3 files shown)


bnagirniak:
Finished adaptive resolution algorithm. Please review.

src/rprblender/engine/viewport_engine.py
```diff
@@ -349,7 +356,17 @@ class FinishRender(Exception):
...
                     # checking for last iteration
                     # preparing information to show in viewport
+                    time_render_prev = time_render
                     time_render = time.perf_counter() - time_begin
+                    iteration_time = time_render - time_render_prev
+                    if depsgraph.scene.rpr.viewport_limits.adapt_viewport_resolution:
+                        if iteration == 2:
+                            target_time = 1.0 / depsgraph.scene.rpr.viewport_limits.viewport_samples_per_sec
+                            render_pixels = self.rpr_context.width * self.rpr_context.height
+                            self.adapt_render_pixels = render_pixels * target_time / iteration_time
```
bsavery, line 366: I think it would be more clear if passing the ratio target_time / iteration_time to adapt_resolution instead of the adapt_render_pixels.

src/rprblender/engine/viewport_engine.py
```diff
@@ -112,31 +112,35 @@ def __init__(self, context: bpy.types.Context):
...
-                self.width = min(render_w, self.width)
-                self.height = int(self.width / region_aspect)
-
     def export_camera(self, rpr_camera):
         """Exports camera settings with render border"""
         self.camera_data.export(rpr_camera,
             ((self.border[0][0] / self.screen_width, self.border[0][1] / self.screen_height),
              (self.border[1][0] / self.screen_width, self.border[1][1] / self.screen_height)))
 
+    def adapt_resolution(self, adapt_render_pixels, prev_settings):
```
bsavery, line 121: I think the names of variables here could be more clear.  
adapt_render_pixels -> requested_num_pixels ?
PIXEL_DIFF -> MIN_PIXEL_DIFF
RATIO_DIFF -> ?

I think this function would be more clear if passing the ratio rather than adapt_render_pixels.  We can get num_pixels desired from ratio * self.width *self.height

src/rprblender/engine/viewport_engine.py
```diff
@@ -112,31 +112,35 @@ def __init__(self, context: bpy.types.Context):
...
+        PIXEL_DIFF = 0.2
+        RATIO_DIFF = 0.1
+        SCALE_MIN = 0.2
```
bsavery, line 124: So if I read this correctly, the viewport will only be scaled to 20% of original size.  How was this setting determined?  Should we have a property for this?

bnagirniak:
In last commit I made cool improvement in viewport rendering by fixing blinking during viewport resizing and changing adaptive resolution. At this stage I think this finished, please review.

src/rprblender/engine/viewport_engine.py
```diff
@@ -112,31 +112,36 @@ def __init__(self, context: bpy.types.Context):
...
-                self.width = min(render_w, self.width)
-                self.height = int(self.width / region_aspect)
-
     def export_camera(self, rpr_camera):
         """Exports camera settings with render border"""
         self.camera_data.export(rpr_camera,
             ((self.border[0][0] / self.screen_width, self.border[0][1] / self.screen_height),
              (self.border[1][0] / self.screen_width, self.border[1][1] / self.screen_height)))
 
+    def adapt_resolution(self, requested_ratio, prev_settings, min_scale):
```
bsavery, line 121: One other thing I noticed here is this can actually scale UP the viewport yes?  I don't think we ever want to do the viewport more than the original resolution.  Might be worth putting in a check for that.

bsavery:
One more thought here I had is if this should be saved with device settings?  Not per scene.

src/rprblender/engine/viewport_engine.py
bnagirniak: Added check that render resoultion <= viewport resolution

bnagirniak:
> 
> 
> One more thought here I had is if this should be saved with device settings? Not per scene.

Made adaptive resolution settings to be saved in user settings.

PR #39, opened 2020-07-23
Per object material export

bsavery:
PURPOSE:  Currently we have a few (3) nodes that need to be exported per object.  There may be more to come

EFFECT:  All materials are exported per object, even if shared.  This may include some time increase, but images are re-used. 

Note:  I thought about setting a "needs_per_object" flag on nodes.  But since the texture coordinate node needs this, I'd claim any complex material will need per object anyway.


src/rprblender/export/material.py
```diff
@@ -23,12 +23,9 @@
...
 def key(material: bpy.types.Material, obj=None, input_socket_key='Surface'):
     mat_key = material.name_full
-    if obj and obj.type == 'MESH' and has_uv_map_node(material):
-        mat_key = (mat_key, obj.data.rpr.uv_sets_names)
-    if input_socket_key != 'Surface':
-        mat_key = (mat_key, input_socket_key)
+    obj_name = obj.name_full if obj is not None else ''
 
-    return mat_key
+    return (mat_key, obj_name, input_socket_key)
```
bnagirniak, line 28: Can be simplified to: return (material.name_full, object.key(obj) if obj else None, input_socket_key)

src/rprblender/engine/viewport_engine.py
```diff
@@ -833,15 +833,11 @@ def sync_collection_instances(self, depsgraph, object_keys_to_export, material_o
...
-        """ Find and remove each unique instance of material by UVs combinations """
-        uvs = set(entry.data.rpr.uv_sets_names for entry in self.depsgraph_objects(depsgraph)
-                  if isinstance(entry, bpy.types.Mesh) and mat.name in entry.material_slots)
-
-        for entry in uvs:
-            for socket in ('Surface', 'Volume', 'Displacement'):
-                mat_key = material.key((mat.name, entry), input_socket_key=socket)
-                if mat_key in self.rpr_context.materials:
-                    self.rpr_context.remove_material(mat_key)
+        """ Find and remove each unique instance of material"""
```
bnagirniak, line 836: Seems remove_material_uvs_instances can be removed at all.

bsavery: I'm not sure what you mean by this comment.

bsavery:
> Left some comments about the code changes.
> But overall, this task is more complicated. Currently it doesn't work properly in viewport, in case when two objects use same material: when I change material, it changes only to one object.

See changes.  I made it so each object with material is updated.

src/rprblender/engine/viewport_engine.py
bnagirniak: I mean, that this function can be removed at all, I checked: it isn't used anywhere now.

bsavery:
BTW in my testing in large scenes, sync times were nearly identical

PR #41, opened 2020-07-27
Add previously unknown vector math modes

bsavery:
EFFECT:  Support Scale, and Trig vector math modes.


src/rprblender/nodes/blender_nodes.py
```diff
@@ -1254,6 +1254,15 @@ def export(self):
...
                 res = abs(in1)
+            elif op == 'SCALE':
+                # input 2 here is a scalar
+                res = in1 * in2
+            elif op == 'SINE':
+                res = in1.sin()
+            elif op == 'COSINE':
+                res = in1.cos()
+            elif op == 'TANGENT':
+                res = in1.tan()
```
bnagirniak, line 1265: One value operations like sin. cos, tan better to move to 1214 line (before else section). Current section relates to 2-values operations.
Also would be good to move there other one-value operations like: FLOOR, CEIL, LENGTH, ABSOLUTE

bsavery:
> Error from Artur
> 
> ```
>   File "C:\Users\artur\AppData\Roaming\Blender Foundation\Blender\2.83\scripts\addons\rprblender\nodes\blender_nodes.py", line 1265, in export
>     res = in1.tan()
> AttributeError: 'NodeItem' object has no attribute 'tan'
> ```

Resolved.

PR #42, opened 2020-07-27
Add support for Magic texture

bsavery:
EFFECT: Magic Texture node is supported now.  By default it uses "generated" texture mapping

Reviewer notes:  I moved the "Generated" Uv function to NodeParser as it is useful in many nodes
Also I noted a bug that 
node_item *= 2 is mistreated but node_item *= 2.0 is ok.  We should handle both.


bsavery:
magic

src/rprblender/nodes/node_parser.py
```diff
@@ -225,6 +225,24 @@ def create_arithmetic(self, op_type, color1, color2=None, color3=None):
 
         return rpr_node
 
+    def get_generated_UV(self):
```
bnagirniak, line 228: More correctly to move this method into NodeParser class from BaseNodeParser, because it uses self.node_item() function which is undefined in BaseNodeParser.

bsavery:
Closing for better implementation in RPR 2.0

PR #43, opened 2020-07-27
Set the camera when exporting .rpr file

bsavery:
EFFECT:  Export engine now uses the scene camera when exporting.


bsavery:
A possible improvement could be to add a setting in the .rpr export dialog.

src/rprblender/engine/export_engine.py
```diff
@@ -69,6 +69,11 @@ def sync(self, context):
         self.rpr_context.set_parameter(pyrpr.CONTEXT_PREVIEW, False)
         scene.rpr.export_ray_depth(self.rpr_context)
 
+        # EXPORT CAMERA
+        camera_key = object.key(scene.camera)   # current camera key
```
ardraeiss, line 73: The scene.camera field can be None if User removes camera from scene or unlink it from this field in UI.
This case causes AttributeError in object key generation.

PR #44, opened 2020-07-27
Better noise texture for 1D and 2D (and an ok option for 3D)

bsavery:
EFFECT:  Better support for 1D and 2D noise.  

Notes: 
Does not support 3D + 4D noise.  These never should have worked.  Perhaps something could be done to combine multiple noise calls in XY, YZ, ZX dimensions to make 3D?


bsavery:
untitled

src/rprblender/nodes/blender_nodes.py
```diff
@@ -1631,27 +1631,136 @@ def export_hybrid(self):
...
+# this is a matrix of offsets prefilled
+# 3 with X offset, 4 with XY offset, 5 with XYZ offset etc.
+# each list of offsets is seeded by the n in length of offsets
+
+RANDOM_OFFSET = []
+for num in range(4):
+    offsets = []
+    for seed in range(num + 3):
+        offset = [0.0, 0.0, 0.0, 0.0]
+        random.seed(seed)
```
ardraeiss, line 1644: Missed "random" module import.

ardraeiss:
Performance is also affected by the number of used arithmetic nodes.

bsavery:
Closing PR.  I am unhappy with performance in RPR 1.0.  RPR 2.0 will have a better solution next week.

PR #45, opened 2020-07-28
RPRBLND-1420: Updating to new RIF 1.5.4

bnagirniak:
PURPOSE
New RIF 1.5.3 is available with some improvements in ML filter using FP16 models.

EFFECT OF CHANGE
Updated to new RIF 1.5.3. Made ML filter to use FP16 models, which could speed up denoising process with less memory usage.

TECHNICAL STEPS
- Switching to latest RIF 1.5.3
- adjusted create_sdk.py and rpr.py
- enabled using FP16 models for ML filter
- added call SynchronizeQueue() before ContextExecuteCommandQueue() which is more correct.

### DIFF
--- a/src/rprblender/engine/image_filter.py
+++ b/src/rprblender/engine/image_filter.py
@@ -102,6 +102,7 @@ class ImageFilter(metaclass=ABCMeta):
             if isinstance(image, rif.FrameBufferImage):
                 image.update()
 
+        self.command_queue.synchronize()
         self.command_queue.execute()
 
     def get_data(self):
--- a/src/rprblender/engine/image_filter.py
+++ b/src/rprblender/engine/image_filter.py
@@ -24,11 +24,6 @@ from rprblender import utils
 from rprblender.utils.user_settings import get_user_settings
 
 
-# Setting environment variable to enable using FP16 models.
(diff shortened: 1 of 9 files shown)


src/rprblender/engine/image_filter.py
```diff
@@ -23,6 +24,11 @@
 from rprblender.utils.user_settings import get_user_settings
 
 
+# Setting environment variable to enable using FP16 models.
+# In next RIF versions with variable won't be needed and can be removed.
+os.environ['RIF_AI_FP16_ENABLED'] = "1"
```
bsavery, line 29: I wonder if we want to support choosing between then.  I'm wondering if FP32 will be more correct now or in future.

bnagirniak: I asked Benjamin about it, he suggested to use FP16. 
Anyway, I heard new RIF will be available this week with some fixes in ML, let’s wait it.

bsavery:
After reviewing I have a few pieces of feedback.  

1.  FP16 for viewport only.  I don't think option should be exposed to users
2.  macOS color-only ML denoiser returns black.  Should be disabled.  I will make these two changes.

bsavery:
Also I am making the ML denoiser default.

PR #47, opened 2020-07-29
RPRBLND-1416: Enable viewport denoising.

bnagirniak:
PURPOSE
Denoiser in viewport only at last iteration isn't enough, would be good to see denoiser results at intermediate iterations.

EFFECT OF CHANGE
Applied denoising in viewport at 4, 8, 16, 32, 64, 96, 128... iterations, not only at last iteration.

TECHNICAL STEPS
Applied denoising in viewport at 4, 8, 16, 32, 64, 96, 128... iterations.

### DIFF
--- a/src/rprblender/engine/viewport_engine.py
+++ b/src/rprblender/engine/viewport_engine.py
@@ -198,8 +198,6 @@ class ViewportEngine(Engine):
         self.is_finished = False
         self.is_synced = False
         self.is_rendered = False
-        self.is_denoised = False
-        self.is_resized = False
 
         self.requested_adapt_ratio = None
         self.is_resolution_adapted = False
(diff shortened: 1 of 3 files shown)


bsavery:
So a couple things here.  
1.  I'm getting a bit of a lag on macOS.  Maybe it needs the FP16 change as well combined with this for optimization?
2.  Can we separate the denoising process from rendering? 
That is, it seems the threads are 
Render 
Render 
Render 
Render
Resolve and denoise (which blocks rendering).  IMO this should not.  Certainly getting the FB needs to block rendering, but denoise part should not block.  Right now we get a bit of a lag at each denoise call.  I would be happier if these denoise calls were more asyncronous.

ardraeiss:
It's better for Artur to check it on different configurations. My PC got some weird errors with denoiser application in viewport for unknown reason.

bnagirniak:
Made another PR #92 for this task. Closing this PR.

PR #48, opened 2020-07-29
RPRBLND-1415: Fix of Adapt viewport resolution.

bnagirniak:
PURPOSE
For some scenes (Mr. elephant for example) adapt viewport resolution algorithm could go into infinite loop of increasing and decreasing render resolution.

EFFECT OF CHANGE
Fixed issue with adap viewport resolution algorithm, when it could go into infinite loop of increasing and decreasing render resolution.

TECHNICAL STEPS
Increased MIN_PIXEL_DIFF to 0.5.

### DIFF
--- a/src/rprblender/engine/viewport_engine.py
+++ b/src/rprblender/engine/viewport_engine.py
@@ -47,8 +47,6 @@ class ViewportSettings:
     """
 
     camera_data: camera.CameraData
-    width: int
-    height: int
     screen_width: int
     screen_height: int
     border: tuple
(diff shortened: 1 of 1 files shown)


bsavery:
Does this mean we need a 50% difference in pixel resolution to adapt to?  Seems like alot

bnagirniak:
Well, I think that our calculation of requested render resolution is quite approximate and rough. Therefore in some scenes (like mr. elephant) 20% of resolution difference isn’t enough. I also tested 40% for this scene and in some camera positions I caught adaptation loop (decreasing/increasing resolution). With 50% it worked ok.

Probably we have to improve algorithm which avoid catching adaptation loop.
I think

bnagirniak:
Algorithm of adapting viewport resolution was rewritten with improvements and fixes:
- fixed adaptation loop, now resolution is adapted only once
- made algorithm more flexible

Please review.

PR #56, opened 2020-08-06
[RPRBLND-1431] Disable anonymous usage statistincs on test and dev machines

ardraeiss:
PURPOSE
Add a way to disable statistics collection for development builds and testing machines.

EFFECT OF CHANGE
- added environment variable RPR_DEV_DISABLE_STATISTICS check to disable anonymous usage statistics collection. Statistics disabled if variable is set to any non-empty value other than "False";
- added config variable disable_athene_report to disable report via configuration.

TECHNICAL STEPS
- added check for RPR_DEV_DISABLE_STATISTICS environment variable presence value before statistics collection;
- added check for config variable disable_athene_report.


bsavery:
Do test machines override config.py?  If so they could use that.

ardraeiss:
Test machines are not overriding config.py at the moment.
Added config option to disable Athena statistics.

PR #58, opened 2020-08-06
Group id update to be consistent with Maya plugin

bsavery:
EFFECT:  Light groups AOVs are numbered 1-4 and choice between 1-4 in light.  Instead of "key" or "fill" options.


src/rprblender/properties/light.py
```diff
@@ -144,10 +144,12 @@ def update_shape(self, context):
     # LIGHT GROUP AOV
     group: EnumProperty(
         name="Light Group",
-        items=(('KEY', "Key", "Key"),
-               ('FILL', "Fill", "Fill")),
+        items=(('1', "1", "Group 1"),
+               ('2', "2", "Group 2"),
+               ('3', "3", "Group 4"),
```
bnagirniak, line 149: “Group 3” should be

src/rprblender/properties/world.py
```diff
@@ -304,6 +304,17 @@ class RPR_EnvironmentProperties(RPR_Properties):
         size=3,
     )
 
+    # LIGHT GROUP AOV
+    group: EnumProperty(
+        name="Light Group",
+        items=(('1', "1", "Group 1"),
+               ('2', "2", "Group 2"),
+               ('3', "3", "Group 4"),
```
bnagirniak, line 312: “Group 3”

bsavery: It's ALMOST like I copy and pasted.  Fixed

PR #59, opened 2020-08-12
[RPRBLND-1419][RPRBLND-1422] Use RPR 2.0 Sphere and Disk lights for Point and Spot

ardraeiss:
PURPOSE
RPR 2.0 added physically correct light sources alternatives to Point Light and Spot Light - the Sphere Light and Disk Light.
Use them when non-zero light size(radius) is used.

EFFECT OF CHANGE
- added Shadow Soft Size to Point light and Spot light UI;
- RPR 2.0 uses more physically correct light render for Point and Spot light when Shadow Soft Size is set more than 0.

TECHNICAL STEPS
- added Shadow Soft Size to Point light and Spot light UI, used existing Blender light property;
- added new light methods to wrapper builder conversion list;
- added RPR 2.0 Sphere and Disk light classes, used in RPR 2.0 mode;
- for compatibility added disk and sphere light placeholders in RPR 1.0 - used spot and point light for that;
- for compatibility added blank set_radius method to RPR 1.0 light base class;
- adjusted Sphere and Disk lights intensity using current light radius to keep the same result intesity;
- adjusted Sphere and Disk lights intensity on radius change.

NOTES
- if light radius is big enough to intersect with illuminated object the dark areas will be present in areas of sphere intersection with said objects;
- Disk light result size changes with Disk light radius change;. softness doesn't change from angle., but raises from radius increase The different method of size calculation, that will account for angle(and visa-verse), is requred.


bsavery:
I have a question.  Shouldn't we use the area light disk and ellipse types to use these instead? They have better convergence than a mesh with emission.

PR #60, opened 2020-08-12
[RPRBLND-1403] Support Vector Math shader node Wrap mode

ardraeiss:
PURPOSE
Vector Math node has "Wrap" mode. Add support for it.

EFFECT OF CHANGE
- added support for "Wrap" mode of Vector Math shader node.

TECHNICAL STEPS
- added 3-inputs operation WRAP support to Vector Math shader node. Adapted Blender util_math.c wrapf method calculations to follow the exact way Blender does it.


src/rprblender/nodes/blender_nodes.py
```diff
@@ -1264,9 +1264,17 @@ def export(self):
...
                 # input 2 here is a scalar
                 res = in1 * in2
-            else:
-                raise ValueError("Incorrect operation", op)
+            else:  # 3-inputs operations
+                in3 = self.get_input_value(2)
 
+                if op == 'WRAP':
+                    val_range = in2 - in3
+                    if range != 0.0:
```
bnagirniak, line 1272: looks like val_range should be here

src/rprblender/nodes/blender_nodes.py
```diff
@@ -1264,9 +1264,17 @@ def export(self):
...
+                        res = in1 - val_range * ((in1 - in3) / val_range).floor()
```
bnagirniak, line 1273: where did you find such formula?

ardraeiss: Cycles

bsavery: Please note so in a comment.  This will explain better for future viewers of code.

bsavery:
PURPOSE
Vector Math node has "Wrap" mode. Add support for it.

EFFECT OF CHANGE

added support for "Wrap" mode of Vector Math shader node.

PR #61, opened 2020-08-13
Updated Northstar libraries for Mac and Win to the latest version

ardraeiss:
PURPOSE
New RPR 2.0 core revision with fixes is ready, use it for testing and integration.

EFFECT OF CHANGE
- used updated RPR 2.0 binaries.

TECHNICAL STEPS
- used updated RPR 2.0 binaries for Mac and Win.


bsavery:
need review from @Dizzyspb  otherwise I approve.

src/rprblender/ui/render.py
```diff
@@ -156,7 +160,9 @@ def draw(self, context):
         col.prop(limits, 'max_samples')
         row = col.row()
         row.prop(limits, 'noise_threshold', slider=True)
-        
+        if context.scene.rpr.render_quality == 'FULL2':
```
bsavery, line 163: min samples should also be disabled technically

PR #62, opened 2020-08-15
Progressive iterations

bsavery:
PROBLEM:  RPR 2.0 has best performance with higher iterations per render call.

EFFECT:  Remove "samples per view update" and replace with a progressively increasing # of samples.  Note that because of the nature of path tracing renders, to get half the noise, we must double the number of samples.  
This PR makes a render use 1,1,2,4,8,16,32,32,32,.... iterations per render call.  This is to give good user feedback and maximize performance.  

Also note that I enabled this in  RPR 1 as well as 2.0.  I would leave as a TODO that preview swatches should just do 32 samples per iteration.


src/rprblender/engine/viewport_engine.py
```diff
@@ -346,7 +352,10 @@ class FinishRender(Exception):
 
                     self.is_rendered = True
                     self.is_denoised = False
-                    iteration += 1
+                    iteration += num_samples
+                    if iteration > 1 and num_samples < 32:
```
ardraeiss, line 356: 32 is just too much for viewport if scene is at least somewhat complex or has refractive objects, even 8 is not that comfortable.

ardraeiss, line 356: Extract the upper limit as a constant

src/rprblender/engine/render_engine.py
```diff
@@ -142,6 +142,9 @@ def _render(self):
                 break
 
             render_iteration += 1
+            if render_iteration > 1 and self.render_update_samples < 32:
```
ardraeiss, line 145: Extract the upper limit as a constant

src/rprblender/engine/render_engine.py
```diff
@@ -242,6 +245,9 @@ def _render_tiles(self):
                     break
 
                 render_iteration += 1
+                if render_iteration > 1 and self.render_update_samples < 32:
```
ardraeiss, line 248: Use the constant extracted from line 145

ardraeiss:
After some testing I propose the new values of top samples step limit:
final render - 16
viewport - 4
The viewport just could not be set to high values as it's whole point is the responsiveness to user actions and higher values are breaking it.

PR #76, opened 2020-08-28
[RPRBLND-1447] Fix Subdivision with RPR2

ardraeiss:
### PURPOSE
Fix the render code to account for RPR 2.0 subdivision application methods.

### EFFECT OF CHANGE
- fixed Subdivision application with RPR 2.0 render quality mode.

### TECHNICAL STEPS
- split subdivision setting method from subdivision parameters calculations;
- used setSubdivisionFactor for RPR2 instead of setAutoAdaptiveSubdivisionFactor.


bnagirniak (approved):
Tested - works ok.

PR #77, opened 2020-08-30
[RPRBLND-952] Unlinked Vector input in procedural texture nodes

ardraeiss:
### PURPOSE
In Cycles the Image Texture node with unlinked Vector(UV) input uses the mesh UV mapping. The procedural texture nodes(such as Noise, Checker, Gradient) are using the Generated UV mapping instead in such case.
The RPR uses UV mapping every time for all supported texture nodes.

To make them more consistent with Cycles result use the Generated UV for supported procedural texture nodes. Leave the Image Texture with UV mapping.

### EFFECT OF CHANGE
- Noise, Checker and Gradient Texture nodes are using Generated UV mapping if Vector input is unlinked.

### TECHNICAL STEPS
- extracted Generated UV mapping coordinates calculations from Texture Coordinates input node as a method;
- used the generated UV mapping for Noise, Checker and Gradient Texture nodes export if Vector input is unlinked.

### NOTES FOR REVIEWERS
- the vertical surfaces have wrong mapping result, just like it was before with Generated UVs;
- the RPR2 doesn't support Generated UV calculations yet.

### DIFF
--- a/src/rprblender/nodes/blender_nodes.py
+++ b/src/rprblender/nodes/blender_nodes.py
@@ -602,9 +602,7 @@ class ShaderNodeTexChecker(NodeParser):
 
         vector = self.get_input_link('Vector')
         if not vector:
-            vector = self.create_node(pyrpr.MATERIAL_NODE_INPUT_LOOKUP, {
-                pyrpr.MATERIAL_INPUT_VALUE: pyrpr.MATERIAL_NODE_LOOKUP_UV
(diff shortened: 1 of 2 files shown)


bnagirniak (commented):
image
Seems for me this is not an expected render result with TexChecker node.

ardraeiss:
Closed until texture application with UVW mapping is fixed

Issue #83, opened 2020-09-06
Error trying to render


gmeter:
Seems like a python error, tried both Linux and Windows with same result.  This is using master branch as well as 2.51

  File "C:\Users\ultra\AppData\Roaming\Blender Foundation\Blender\2.90\scripts\addons\rprblender\nodes\blender_nodes.py", line 1637, in <genexpr>
    in_col = self.get_input_scalar('Color')
TypeError: 'float' object is not subscriptable

bsavery:
Can you attach the full error stack?

bsavery:
Or better yet a scene to recreate this

gmeter:
I put it on a google Drive, link below.  I embedded the textures incase they were the problem.
Also not that this is using Blender version 8.90

https://drive.google.com/file/d/16PCRurtx9OPvWGxAXF-fBb1dHJzZT0b9/view?usp=sharing

bsavery:
Hi.  I tried this both in RPR 2.0 and Full (RPR 1.0) rendering and can't see this error.  Can you go to the render settings -> About section and check your version number?

Also what render mode are you using?  Full?

bnagirniak:
I checked - it is repro for me on Blender 2.90, I can fix it.

gmeter:
I can confirm that bnagirniak's fix solves the problem

PR #88, opened 2020-09-11
RPRBLND-1471: Possible error in parsing ShaderNodeRGBCurve.

bnagirniak:
### PURPOSE
Possible error in parsing ShaderNodeRGBCurve, more details #83.
`TypeError: 'float' object is not subscriptable`

### EFFECT OF CHANGE
Fixed possible error in parsing RGB Curve node.

### TECHNICAL STEPS
Fixed by using NodeItem.get_channel(), which works with float data value.


ardraeiss (approved):
Looks OK

PR #92, opened 2020-09-22
RPRBLND-1416: Enable viewport denoising

bnagirniak:
### PURPOSE
Denoiser in viewport only at last iteration isn't enough, would be good to see denoiser results at intermediate iterations.

### EFFECT OF CHANGE
Applied denoising in viewport at 4, 8, 16, 32, 64, 96, 128... iterations, not only at last iteration. Increased viewport speed.

### TECHNICAL STEPS
1. Applied denoising in viewport at 4, 8, 16, 32, 64, 96, 128... iterations.
2. Improved GLTexture, now it holds image pixels and can be resized during set_image().
3. Increased viewport speed by: 
    - with enabled denosier: denosied image is prepared once and is used to draw without getting data from image filter + resolve context doesn't called
    - with disabled denoiser: resolve context is called in draw() so it could be called not every iteration.

### NOTES FOR REVIEWERS
Denoising is applied in the same thread were rendering is applied. Due to denoiser is applied only at 4, 8, 16, 32, 64, 96, 128... iterations it doesn't so noticeable especially on heavy scenes.
I tried several approaches and did lot of investigation of were denosier should be applied: 
1. In render thread - most stable and simple algorithm. Decided to use this.
2. In separate denoiser thread - slightly faster then 1) but less stable during resizing and other scene updates.
3. In draw() method - quite stable but could hung viewport, speed is same as 1).

### DIFF
--- a/src/rprblender/engine/viewport_engine.py
+++ b/src/rprblender/engine/viewport_engine.py
@@ -186,7 +186,7 @@ class ViewportEngine(Engine):
     def __init__(self, rpr_engine):
         super().__init__(rpr_engine)
 
-        self.gl_texture: gl.GLTexture = None
+        self.gl_texture = gl.GLTexture()
         self.viewport_settings: ViewportSettings = None
         self.world_settings: world.WorldData = None
         self.shading_data: ShadingData = None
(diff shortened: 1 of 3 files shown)


ardraeiss (approved):
Looks OK

PR #94, opened 2020-09-23
Rename blender 28x exe build variable.

bsavery:
### PURPOSE
Rename build methods using BLENDER_28x_EXE to BLENDER_EXE.  Since we can now build with 2.90 etc.

### EFFECT OF CHANGE
No functional change

### TECHNICAL STEPS
DEVELOPERS build scripts may need to be updated

### DIFF
--- a/README-OSX.md
+++ b/README-OSX.md
@@ -60,10 +60,10 @@ NOTES:
 
 ### Running
 
-Ensure that the BLENDER_28X_EXE environment variable is set.  For example, add the following to
+Ensure that the BLENDER_EXE environment variable is set.  For example, add the following to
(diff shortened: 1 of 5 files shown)


ardraeiss (approved):

bnagirniak (approved):

bnagirniak:
What about merging this PR into develop?

bsavery:
> What about merging this PR into develop?

Sorry forgot about this.  Would it affect CIS builds?

bnagirniak:
It shouldn't because BLENDER_EXE is used only for local debug running plugin, but CIS runs only installed plugin.

ardraeiss:
It affects debug mode launch CMD/SH scripts only.

PR #95, opened 2020-09-23
Material Index and fix Object Index

bsavery:
### PURPOSE
Object Index is now fully supported in core, as is material index

### EFFECT OF CHANGE
Material index "Pass Index" and Object Index now fully match cycles.

### TECHNICAL STEPS
Note that the lookup table must be set via set_aov_index_lookup()


src/rprblender/engine/engine.py
```diff
@@ -96,11 +96,6 @@ def zeros_image(channels):
             if p.channels != image.shape[2]:
                 image = image[:, :, 0:p.channels]
 
-            if p.name == "Object Index":
```
bsavery, line 99: No longer needed with lookup table.

bnagirniak (approved):
Fixed export in RPR 1.0 by moving set_aov_index_lookup to RPRContext. Made small cleanups.

src/rprblender/export/material.py
```diff
@@ -90,7 +90,6 @@ def sync(rpr_context: RPRContext, material: bpy.types.Material, input_socket_key
     rpr_material = node_parser.final_export(input_socket_key)
 
     if rpr_material:
-        rpr_context.set_material_node_as_material(mat_key, rpr_material)
```
bsavery, line 93: @bnagirniak what is the reason for this?

src/rprblender/ui/material.py
```diff
@@ -122,6 +122,9 @@ class RPR_MATERIAL_PT_surface(RPR_MaterialOutputSocket):
     def draw(self, context):
         layout = self.layout
 
+        row = layout.row()
+        row.prop(context.material, 'pass_index')
```
ardraeiss, line 126: I'm not sure it's needed here.
This parameter is already present in Material tab -> Viewport Display section ->Settings subsection, enabled in RPR UI.

ardraeiss (approved):
Code seems to be fine.

There are issues with results:
- the non-zero Material Index values are correct in RPR 2.0;
- Material Index value of "0" has wrong result in RPR 2.0. I've got "0.403" and "3.0" for two separate "0" materials in test scene;
- any Material Index result is wrong in RPR 1.0.

src/rprblender/export/material.py
bnagirniak: I removed this line because it is already exist below, It was duplicated in this block.

bsavery:
> * Material Index value of "0" has wrong result in RPR 2.0. I've got "0.403" and "3.0" for two separate "0" materials in test scene;

Perhaps should only set if index is != 0

> * any Material Index result is wrong in RPR 1.0.

RPR 1.0 is considered legacy at this point.  Won't fix.

src/rprblender/ui/material.py
bsavery: Hmmm ok, I didn't see it in there, kinda hidden.  But I'll remove.

ardraeiss:
The Cycles threads 0 as "0" in Materia ID AOV output. Does RPR uses 0 value as an "Auto" mode flag? Not sure it can be fixed from plugin side.

bsavery:
I think there was a missed line in the checking with the 0 issue.

PR #97, opened 2020-09-24
Skipped zero-sized area lights

ardraeiss:
### PURPOSE
Zero-sized(light area X*Y results in 0) raises RuntimeWarning of "divide by zero". It doesn't break the render thanks to the Blender handling it but it's still a bug. Fix it.

### EFFECT OF CHANGE
- area lights with 0 area are ignored.

### TECHNICAL STEPS
- added check for area light area. Zero-sized lights are skippedж
- made sure the Area light area is positive number or zero.

### NOTES
-  Cycles result is black for such light source. That's why skipping light creation is used, to follow Cycles result.


bsavery:
IMO instead of skipping why not do point light?

ardraeiss:
Cycles result is black for zero-sized Area light. That's why skipping.

bsavery:
> Cycles result is black for zero-sized Area light. That's why skipping - to follow Cycles results.

Fair enough.

src/rprblender/export/light.py
```diff
@@ -146,6 +146,8 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, instance_key=None):
     elif light.type == 'AREA':
         data = mesh.MeshData.init_from_shape_type(rpr.shape, light.size, light.size_y, segments=32)
         area = data.area * obj.scale[0] * obj.scale[1]
```
bnagirniak, line 148: Please provide one more fix here:
area = abs(data.area * obj.scale[0] * obj.scale[1])
because obj.scale could be below zero.

bnagirniak (commented):
Suggested one more fix here.

bnagirniak (approved):

PR #100, opened 2020-10-01
Explicitly removed render update callback on render end

ardraeiss:
### PURPOSE
With render update callback introduction the VRAM consumption of RPR 2.0 was increased. The memory was held for some time between renders. It caused performance spikes, sometimes crashes if memory wasn't freed in time.

### EFFECT OF CHANGE
- fixed performance and VRAM consumption issue in RPR 2.0 mode.

### TECHNICAL STEPS
- explicitly removed render update callback on render end;
- removed previous callback removal method since it doesn't worked as expected.


bnagirniak (changes requested):
Add explicit remove callback from ViewportEngine2

bnagirniak (approved):
Tested - works good.

PR #110, opened 2020-10-28
Make RPR 2 default

bsavery:
### PURPOSE
Make RPR 2 default renderer.  And change naming of Full and Full2

### EFFECT OF CHANGE
User gets RPR2 by default (named Full).  Old "Full" mode renamed to Legacy mode.


src/rprblender/properties/render.py
```diff
@@ -429,7 +429,7 @@ def is_tile_render_available(self):
         ]
     if pyrpr2.enabled:
         render_quality_items += [
-            ('FULL2', "RPR 2.0 (Beta)", "Full render quality with RPR 2 (Beta)")
+            ('FULL2', "Full render quality using RPR 2, including hardware ray tracing support.")
```
bnagirniak, line 432: Name of option is missed here, which raises: ValueError: bpy_struct "RPR_RenderProperties" registration error: render_quality could not register

src/rprblender/properties/render.py
```diff
@@ -444,7 +444,7 @@ def update_render_quality(self, context):
         name="Render Quality",
         description="RPR render quality",
         items=render_quality_items,
-        default='FULL',
+        default='FULL2' if pyrpr2.enabled else 'FULL',
```
bnagirniak, line 447: What do you think about removing pyrpr2.enabled and config.enable_rpr2 options at all in this PR?

bnagirniak (changes requested):
Error in render_quality_items and proposition to remove pyrpr2.enabled option.

src/rprblender/properties/render.py
```diff
@@ -429,7 +429,7 @@ def is_tile_render_available(self):
         ]
     if pyrpr2.enabled:
```
ardraeiss, line 430: With config variable removal this check produces error since pyrpr2.enabled is False.
It seems this check should be removed.

ardraeiss (changes requested):

ardraeiss (approved):
OK

src/rprblender/properties/render.py
```diff
@@ -419,18 +419,17 @@ def is_tile_render_available(self):
...
             ('HIGH', "High", "High render quality"),
             ('MEDIUM', "Medium", "Medium render quality"),
             ('LOW', "Low", "Low render quality"),
         ]
-    if pyrpr2.enabled:
-        render_quality_items += [
-            ('FULL2', "RPR 2.0 (Beta)", "Full render quality with RPR 2 (Beta)")
-        ]
+    render_quality_items += [
+        ('FULL2', "Full", "Full render quality using RPR 2, including hardware ray tracing support.")
```
bnagirniak, line 431: In UI 'Full' quality is last item, better to have it as first. Would be good to move this item to 422 line.

bnagirniak (commented):
One suggestion about quality items order.

PR #112, opened 2020-10-29
RPRBLND-1520: Move to use Direct transform setting for motion blur

bnagirniak:
### PURPOSE
Motion blur functionality for RPR 2.0 was integrated in new core with rprShapeSetMotionTransform() and rprShapeSetMotionTransformCount() functions. 

### EFFECT OF CHANGE
Implemented motion blur for RPR 2.0.

### TECHNICAL STEPS
- Implemented pyrpr2.Shape.set_motion_transform(), pyrpr2.AreaLight.set_motion_transform().
- Adjusted Engine.sync_motion_blur() and particle.sync() to use new functionality.

### NOTES FOR REVIEWERS
RPR 1 still uses old functionality.

### DIFF
--- a/src/bindings/pyrpr/src/pyrpr2.py
+++ b/src/bindings/pyrpr/src/pyrpr2.py
@@ -39,6 +39,25 @@ class Context(pyrpr.Context):
             self.render_update_callback = None
 
 
+class Shape(pyrpr.Shape):
+    def set_motion_transform(self, transform, transpose=True): # Blender needs matrix to be transposed
+        pyrpr.ShapeSetMotionTransformCount(self, 1)
+        pyrpr.ShapeSetMotionTransform(self, transpose, pyrpr.ffi.cast('float*', transform.ctypes.data), 1)
+
+
+class Mesh(pyrpr.Mesh, Shape):
+    pass
+
+
+class Instance(pyrpr.Instance, Shape):
+    pass
+
+
+class AreaLight(pyrpr.AreaLight):
+    def set_motion_transform(self, transform, transpose=True): # Blender needs matrix to be transposed
+        self.mesh.set_motion_transform(transform, transpose)
+
+
 class SphereLight(pyrpr.Light):
     def __init__(self, context):
         super().__init__(context)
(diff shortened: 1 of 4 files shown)


src/bindings/pyrpr/src/pyrpr2.py
```diff
@@ -39,6 +39,25 @@ def render_update_callback(progress, data):
             self.render_update_callback = None
 
 
+class Shape(pyrpr.Shape):
+    def set_motion_transform(self, transform, transpose=True): # Blender needs matrix to be transposed
```
bsavery, line 43: I think this is ok for now. If for the future there was support for > 2 motion transforms this could be more generic like:

set_motion_transform(self, transform, time, transpose=True)

src/rprblender/engine/engine.py
```diff
@@ -126,23 +126,26 @@ def depsgraph_instances(self, depsgraph: bpy.types.Depsgraph):
                 yield instance
 
     def sync_motion_blur(self, depsgraph: bpy.types.Depsgraph):
-        """ """
 
         def set_motion_blur(rpr_object, prev_matrix, cur_matrix):
-            velocity = (prev_matrix - cur_matrix).to_translation()
-            rpr_object.set_linear_motion(*velocity)
+            if hasattr(rpr_object, 'set_motion_transform'):
```
bsavery, line 131: should this just be "if pyrpr2.enabled" or something.

src/bindings/pyrpr/src/pyrpr2.py
bnagirniak: yes

src/rprblender/engine/engine.py
bnagirniak: Another option here to use: if isinstance(rpr_object, (pyrpr2.Shape, pyrpr2.AreaLight)).

bsavery: That seems less better.  aka worse

src/bindings/pyrpr/src/pyrpr2.py
bsavery: From core team this might come sooner rather than later, not sure if you want to make this change now.

ardraeiss (approved):
Blur is fine with these changes

src/bindings/pyrpr/src/pyrpr2.py
bnagirniak: I'd prefer to make this change, when set_motion_transform be available in RPR1. Anyway this is very easy change.

PR #114, opened 2020-10-29
Create setting for Texture Cache dir and operator to clear.

bsavery:
### PURPOSE
Expose RPR2 Texture Cache directory

### EFFECT OF CHANGE
User has a setting in debug panel for setting texture cache dir (default to .tex_cache in addon).  User can change or clear the dir.


src/rprblender/properties/render.py
```diff
@@ -493,6 +499,11 @@ def init_rpr_context(self, rpr_context, is_final_engine=True, use_gl_interop=Fal
             if float(mac_vers_major) >= 14:
                 rpr_context.set_parameter(pyrpr.CONTEXT_METAL_PERFORMANCE_SHADER, 1)
 
+        # enable texture cache
+        if not os.path.isdir(self.texture_cache_dir):
```
ardraeiss, line 503: There could be file with such name, the os.path.isdir will return False. The mkdir will fail in this case.

bsavery: eh.... I mean that's a pretty weird situation.  The file dir selector doesn't let you choose a single file.

src/rprblender/operators/render.py
```diff
@@ -46,6 +46,21 @@ def execute(self, context):
...
+    '''
+    Clear texture cache dir
+    '''
+
+    bl_idname = "rpr.op_clear_tex_cache"
+    bl_label = "Clear Cache"
+    bl_description = "Clear Texture Cache"
+
+    def execute(self, context):
+        shutil.rmtree(context.scene.rpr.texture_cache_dir)
```
bnagirniak, line 60: shutil is not imported, ImportError here.

src/rprblender/properties/render.py
```diff
@@ -493,6 +499,11 @@ def init_rpr_context(self, rpr_context, is_final_engine=True, use_gl_interop=Fal
...
+            os.mkdir(self.texture_cache_dir)
+        rpr_context.set_parameter(pyrpr.CONTEXT_TEXTURE_CACHE_PATH, self.texture_cache_dir)
```
bnagirniak, line 505: CONTEXT_TEXTURE_CACHE_PATH is available only for RPR 2, please add check isinstance(rpr_context, RPRContext2) or override RPRContext2.set_parameter() with ignoring CONTEXT_TEXTURE_CACHE_PATH option.

bnagirniak (changes requested):
Found bugs + suggestion using utils.package_root_dir().

src/rprblender/properties/render.py
```diff
@@ -493,6 +499,12 @@ def init_rpr_context(self, rpr_context, is_final_engine=True, use_gl_interop=Fal
             if float(mac_vers_major) >= 14:
                 rpr_context.set_parameter(pyrpr.CONTEXT_METAL_PERFORMANCE_SHADER, 1)
 
+        # enable texture cache for RPR2
+        if isinstance(rpr_context, pyrpr2.Context):
```
bnagirniak, line 503: should be:
`if isinstance(rpr_context, RPRContext2):`

bnagirniak (changes requested):
Found bug in checking rpr_context type

bnagirniak (approved):

ardraeiss (approved):

PR #115, opened 2020-10-30
Render mode in final render

bnagirniak:
### PURPOSE
Render Mode option can be added for final render too.

### EFFECT OF CHANGE
Added Render Mode option to final render, updated UI by disabling it for hybrid, as not supported.

### TECHNICAL STEPS
- added scene.rpr.export_render_mode to final render sync()
- updated ui/render.py and ui.view3d.py with render_mode option which is disabled for hybrid.

### DIFF
--- a/src/rprblender/engine/render_engine.py
+++ b/src/rprblender/engine/render_engine.py
@@ -469,6 +469,7 @@ class RenderEngine(Engine):
 
         # SET rpr_context parameters
         self.rpr_context.set_parameter(pyrpr.CONTEXT_PREVIEW, False)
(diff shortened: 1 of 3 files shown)


bnagirniak:
I initially misunderstand this task and made something else here. Contour rendering is a new feature in new Core, it'll be implemented in another PR.
Do we need changes which were made here?

ardraeiss:
> Do we need changes which were made here?

It seems useful, rename the PR to reflect what it actually does.

bnagirniak:
Looks like such changes aren't needed, closing this PR.

PR #119, opened 2020-11-03
RPRBLND-1531: Implement 'preview' setting in viewport for RPR2

bnagirniak:
### PURPOSE
'Preview' option was implemented in latest core 2.0.1.7-final and viewport can be rendered more faster at first iterations. Therefore it make sense to show render results of first iterations during viewport changes.

### EFFECT OF CHANGE
Implemented showing render results of first iterations during viewport changes for RPR 2.0.

### TECHNICAL STEPS
Done:
- set render update callback starting from second iteration
- made getting render result for first iteration in _do_render thread.
- switched RadeonProRenderSDK to 2.0.1.7 final

### DIFF
--- a/src/rprblender/engine/viewport_engine_2.py
+++ b/src/rprblender/engine/viewport_engine_2.py
@@ -62,7 +62,7 @@ class ViewportEngine2(ViewportEngine):
 
             # don't need to do intermediate update for 0, 1 iteration and
             # at render finish when progress == 1.0
-            if iteration <= 1 or progress == 1.0:
+            if progress == 1.0:
                 return
 
             self.resolve_event.set()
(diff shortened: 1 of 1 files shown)


ardraeiss (commented):
It's fine.
The new core version approval from QA is required for final approval.

bsavery (changes requested):
Its unclear to me about which part is exposing the northstar preview setting?  Also should we be setting preview setting > 1?

bnagirniak:
> 
> 
> Its unclear to me about which part is exposing the northstar preview setting? Also should we be setting preview setting > 1?

Yes, such new feature with preview setting > 1 wasn't implemented and was new for me. Now it is integrated.
I also played with different values and combinations of 'preview' and found the best choice is 3 for first iteration and 0 (disabling) for next iterations.

PR #120, opened 2020-11-03
Set zip compression level to Z_BEST_COMPRESSION in create_zip_addon.py.

bnagirniak:
### PURPOSE
We didn't use any zip compression level for zipping plugin installer. Therefore size of plugin took 166MB in Windows. With zip compression the size decreases to 103 MB.

### EFFECT OF CHANGE
Decreased zip installer plugin size by 40% approximately.

### TECHNICAL STEPS
Set compression=zipfile.ZIP_DEFLATED and compresslevel=zlib.Z_BEST_COMPRESSION in creating ZipFile.

### DIFF
--- a/BlenderPkg/create_zip_addon.py
+++ b/BlenderPkg/create_zip_addon.py
@@ -14,6 +14,7 @@
 #********************************************************************
 
 import zipfile
+import zlib
 import platform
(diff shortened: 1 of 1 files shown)


ardraeiss (approved):
Code is fine

PR #122, opened 2020-11-05
RPRBLND-1502: Memory leak with specific scene

bnagirniak:
### PURPOSE
This is an old bug (from beginning), which is reproducible on scenes which have material with lot of arithmetic. Memory leak especially noticeable during animation render.

### EFFECT OF CHANGE
Fixed memory leak during rendering some scenes, which was especially noticeable during animation render.

### TECHNICAL STEPS
Done:
- added pyrpr.MaterialNode.delete() which removes all links from material node and deletes node more correctly;
- adjusted pyhybrid.MaterialNode.delete() which just removes links;
- small cleanup in pyrpr.Object.

### DIFF
--- a/src/bindings/pyrpr/src/pyhybrid.py
+++ b/src/bindings/pyrpr/src/pyhybrid.py
@@ -143,6 +143,10 @@ class MaterialNode(pyrpr.MaterialNode):
 
         super().set_input(name, value)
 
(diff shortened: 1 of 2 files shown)


bsavery (approved):

ardraeiss (approved):
I like it

PR #124, opened 2020-11-06
[RPRBLND-1519] Contour render mode

ardraeiss:
### PURPOSE
RPR2 supports new Contour render mode. Use it.

### EFFECT OF CHANGE
- added new Contour render mode for the final render.

### TECHNICAL STEPS
- added Contour mode properties to render properties;
- added Contour mode UI to Render tab;
- added object Contour mode visibility flag to Visibility settings panel;
- set RPR context GPU integrator to "gpucontour" if Contour mode enabled;
- set Contour context parameters if Contour mode enabled;
- added use_contour flag to objects/instances export to correctly exclude excluded objects from Contour results;
- added default texture cache .tex_cache folder to .gitignore;
- set Contour to start render from 4 iterations.

### NOTES FOR REVIEWERS
Viewport in Contour mode is slow and ugly, additional work is required. Excluded from this pull request.


bnagirniak (approved):
Reviewed and tested - works ok.

src/rprblender/properties/object.py
```diff
@@ -73,6 +73,11 @@ class RPR_ObjectProperites(RPR_Properties):
         description="Use this object as a portal light",
         default=False,
     )
+    visibility_contour: BoolProperty(
```
bsavery, line 76: What would be the reason you'd want an object to be invisible in contour mode?

src/rprblender/ui/render.py
```diff
@@ -219,6 +219,48 @@ def draw(self, context):
         self.layout.prop(rpr_scene, 'ray_cast_epsilon', slider=True)
 
 
+class RPR_RENDER_PT_contour(RPR_Panel):
```
bsavery, line 222: Please rename "contour rendering"

src/rprblender/properties/object.py
ardraeiss: It is the part of Parameters and UI sections of Contour.pptx attached to the task

bsavery: Ok.  I'll talk to SDK team.  This is a weird setting it should just use camera visibility.

PR #127, opened 2020-11-11
RPRBLND-1538: Low subdivision level causes a system crash.

bnagirniak:
### PURPOSE
There are several issues in subdivision settings:
- low subdivision level causes a system crash;
- default subdivision level as 1.0 subdivides mesh so much, which to noticeably slows whole rendering;
- have several blames about misunderstanding of what subdivision level means and why it has values from 0.01 to 10.0 (soft max);
- RPR1 and RPR2 produce so different mesh subdivision in the same subdivision level.

### EFFECT OF CHANGE
Improved working with subdivision settings: changed behavior of subdivision level as integer value from 0 to 8 (as soft max) and 12 (as max).

### TECHNICAL STEPS
- changed behavior of subdivision level as integer value from 0 to 8 (as soft max) and 12 (as max)
- adjusted levels between RPR 1 and RPR 2
- made some code improvements

### DIFF
--- a/src/rprblender/properties/object.py
+++ b/src/rprblender/properties/object.py
@@ -96,7 +96,7 @@ class RPR_ObjectProperites(RPR_Properties):
         name="Level",
         description="Subdivision level for mesh. For finer subdivision set upper",
         min=0, max=12, soft_max=8,
-        default=1
+        default=3
     )
     subdivision_boundary_type: EnumProperty(
         name="Boundary Type",
--- a/src/rprblender/properties/object.py
+++ b/src/rprblender/properties/object.py
@@ -92,9 +92,16 @@ class RPR_ObjectProperites(RPR_Properties):
         description="Enable subdivision",
         default=False,
     )
-    subdivision_factor: IntProperty(
+    subdivision_factor: FloatProperty(
+        name="Polygon Size",
+        description="Subdivision polygon size, in pixels that it should be subdivided to.\n"
+                    "For finer subdivision set lower.",
+        min=0.5, soft_max=512.0,
+        default=16.0
+    )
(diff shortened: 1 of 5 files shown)


src/rprblender/properties/object.py
```diff
@@ -92,11 +92,11 @@ class RPR_ObjectProperites(RPR_Properties):
...
-    subdivision_factor: FloatProperty(
-        name="Adaptive Level",
-        description="Subdivision factor for mesh, in pixels that it should be subdivided to. For finer subdivision set lower.",
-        min=0.01, soft_max=10.0,
-        default=1.0
+    subdivision_factor: IntProperty(
+        name="Level",
+        description="Subdivision level for mesh. For finer subdivision set upper",
+        min=0, max=12, soft_max=8,
+        default=1
```
ardraeiss, line 99: Min level could be risen to 4 to keep default behavior like before.

ardraeiss (approved):
It's okey.

ardraeiss (approved):

src/rprblender/properties/object.py
```diff
@@ -93,10 +93,17 @@ class RPR_ObjectProperites(RPR_Properties):
         default=False,
     )
     subdivision_factor: FloatProperty(
-        name="Adaptive Level",
-        description="Subdivision factor for mesh, in pixels that it should be subdivided to. For finer subdivision set lower.",
-        min=0.01, soft_max=10.0,
-        default=1.0
+        name="Polygon Size",
```
bsavery, line 96: Change to "Subdiv Polygon size"

bsavery (commented):
Just a change on name of prop

PR #129, opened 2020-11-12
[RPBLND-1547] Incorrect RPR Bump Map node result

ardraeiss:
### PURPOSE
The RPR Bump Map node is used by some Material Library materials. The result is different from Blender Bump shader node. With new core 2.1.8 introduction this issue became even more apparent, the result is just flat in RPR2 mode.
Fix it.

### EFFECT OF CHANGE
- fixed incorrect RPR Bump Map shader node result.

### TECHNICAL STEPS
- updated RPR Bump Map node to work in the same way Blender Bump node does: blend(normal, normal+bump, scale).

### NOTES FOR REVIEWERS
Bug is more noticeable with core 2.1.8.


bnagirniak (approved):
Looks good

PR #130, opened 2020-11-12
Use OCIO texture colorspace

bsavery:
### PURPOSE
Blender includes a config.ocio file for texture color spaces.  We can use this to define the color space for textures and set the ocio colorspace for image textures for RPR2

### EFFECT OF CHANGE
Texture Colorspace is respected better in RPR2, particularly for non-srgb or raw color spaces

### NOTES FOR REVIEWERS
A further change could be to use the OCIO render space to use the Filmic setting, not sure if this would help us any.
Also Blender actually allows an environment variable to do a custom config.ocio file. That's not supported but could be.


src/rprblender/export/image.py
```diff
@@ -94,9 +95,10 @@ def sync(rpr_context, image: bpy.types.Image, use_color_space=None, frame_number
 
     rpr_image.set_name(str(image_key))
 
-    # TODO: implement more correct support of image color space types
     # RPRImageTexture node color space names are in caps, unlike in Blender
-    if color_space in ('sRGB', 'BD16', 'Filmic Log', 'SRGB'):
+    if isinstance(rpr_context, context.RPRContext2):
```
ardraeiss, line 99: How does it work with images that should be used in "Linear"/"Raw" color space only - Normal maps, roughness maps, ets?

src/rprblender/properties/render.py
```diff
@@ -580,6 +580,12 @@ def init_rpr_context(self, rpr_context, is_final_engine=True, use_gl_interop=Fal
             if float(mac_vers_major) >= 14:
                 rpr_context.set_parameter(pyrpr.CONTEXT_METAL_PERFORMANCE_SHADER, 1)
 
+            # set ocio config file to blender included one
+            # TODO can use blender render set render space, and check for custom ocio setting
+            rpr_context.set_parameter(pyrpr.CONTEXT_OCIO_CONFIG_PATH,
```
bnagirniak, line 585: Why do we set it under metal_enabled (in MacOS) option? Is this relates only to MacOS?
Also for RPR1 it produces raises ERROR_UNSUPPORTED.

bnagirniak (changes requested):
I tested this PR on Windows and effect was like it just disables texture gamma. What effect should we expect here?
Also found some errors, which were commented.

src/rprblender/properties/render.py
bsavery: This was a copy and paste error from another branch.  Will fix.

src/rprblender/export/image.py
bsavery: fine.  but just like blender and cycles the user has to set that manually.

bnagirniak (commented):
Now it works ok, but it changes so much overall behavior, because it doesn't apply gamma to textures. All scenes became to look so different in RPR2 comparing to RPR1.
Maybe some UI checkbox is needed for this changes?

bsavery:
Converting to draft for now.

bsavery:
> Now it works ok, but it changes so much overall behavior, because it doesn't apply gamma to textures. All scenes became to look so different in RPR2 comparing to RPR1.
> Maybe some UI checkbox is needed for this changes?

This maybe needs a bit of debugging.  The idea of this is it should be applying the srgb based on OCIO

bsavery:
Rebased to develop.  This SHOULD work now.  I'm not sure what's going on in the core if it's not.

bnagirniak:
Looks like this PR includes lot of previous commits and with lot of changes. Can you please create a new branch basing on develop branch and cherry pick there only necessary commits?

PR #131, opened 2020-11-16
RPRBLND-1555: Viewport 'preview' rendering produces square artifacts at next iterations

bnagirniak:
### PURPOSE
Using 'preview' rendering mode in viewport produces square artifacts at next iterations.

### EFFECT OF CHANGE
Fixed viewport rendering artifacts after using 'preview' rendering mode.

### TECHNICAL STEPS
Fixed by clearing framebuffers before second iteration disabling intermediate update at second iteration.


ardraeiss (approved):
Now viewport looks much better

PR #142, opened 2020-11-20
Changes to match cycles default

bsavery:
### PURPOSE
To match cycles lighting better with default settings
I identified 3 places we were very different
1.  Cycles ray visibility settings 
2.  Point, spot, directional intensity was too high (spot and point seem to be ^2 of cycles)
3.  Directional light softness
4.  Velvet shader

### EFFECT OF CHANGE
"Sun" light softness should match Cycles better
Default intensity of Sun, spot, point light matches cycles better
Cycles Visibility settings are for RPR
Clamp radiance is disabled by default.  Users can enable to kill fireflies.
Velvet shader looks closer by default to Cycles


src/rprblender/export/light.py
```diff
@@ -45,6 +45,7 @@ def get_radiant_power(light: bpy.types.Light, area=0.0):
     if light.type in ('POINT', 'SPOT'):
         units = rpr.intensity_units_point
         if units == 'DEFAULT':
+            default_intensity = color * math.sqrt(light.energy) # to match cycles
```
ardraeiss, line 48: Way too dark if combined with the next line. Use
return color * math.sqrt(light.energy)
instead and remove the next line.

src/rprblender/nodes/blender_nodes.py
```diff
@@ -410,20 +410,28 @@ class ShaderNodeBsdfVelvet(RuleNodeParser):
...
+                pyrpr.MATERIAL_INPUT_COLOR1: "inputs.Sigma",
+            }
+        },
+
         "BSDF": {
             "type": pyrpr.MATERIAL_NODE_UBERV2,
             "params": {
                 pyrpr.MATERIAL_INPUT_UBER_DIFFUSE_COLOR: "inputs.Color",
-                pyrpr.MATERIAL_INPUT_UBER_DIFFUSE_WEIGHT: "inputs.Sigma",
+                pyrpr.MATERIAL_INPUT_UBER_DIFFUSE_WEIGHT: "nodes.ONE_MINUS_SIGMA",
```
ardraeiss, line 425: Sigma can be set to < 0.0. The Cycles result is just black if sigma 0.0 and less, this change makes RPR way too bright.

src/rprblender/export/mesh.py
```diff
@@ -270,6 +270,10 @@ def sync_visibility(rpr_context, obj: bpy.types.Object, rpr_shape: pyrpr.Shape,
     obj.rpr.export_visibility(rpr_shape, indirect_only)
     obj.rpr.export_subdivision(rpr_shape)
 
+    if is_emissive_object(obj):
```
ardraeiss, line 273: As we've discussed - not needed, it was the effect of Ray Visibility settings difference between RPR and Cycles. Separate PR for that ray visibility changes would be better.

ardraeiss (changes requested):

src/rprblender/nodes/blender_nodes.py
bsavery: I would claim those are invalid values.  At any rate.  My goal here was to make BMW floor make good.

src/rprblender/ui/object.py
```diff
@@ -48,13 +48,14 @@ def draw(self, context):
...
+        cycles_vis = context.object.cycles_visibility
 
         flow = self.layout.grid_flow(row_major=True, even_columns=True)
-        flow.column().prop(rpr, 'visibility_in_primary_rays')
-        flow.column().prop(rpr, 'reflection_visibility')
-        flow.column().prop(rpr, 'refraction_visibility')
-        flow.column().prop(rpr, 'diffuse_visibility')
-        flow.column().prop(rpr, 'shadows')
+        flow.column().prop(cycles_vis, 'camera')
+        flow.column().prop(cycles_vis, 'reflection')
```
ardraeiss, line 55: Wrong Cycles property name, should be:
flow.column().prop(cycles_vis, 'glossy')

src/rprblender/nodes/blender_nodes.py
ardraeiss: Cycles does clamp(0.0, 1.0) on Sigma input value. I think so should we.

ardraeiss (changes requested):
Small issue with rays visibility UI

ARozghon:
"Wanderer" scene from Blender cloud scene has a lot of needed nodes in "World" Shader editor settings

bnagirniak (commented):
Overall - looks ok, tested from my side. Waiting for fix @ardraeiss mentioned with flow.column().prop(cycles_vis, 'glossy')

PR #144, opened 2020-11-24
[RPRBLND-1556] Limit Contour iterations number

ardraeiss:
### PURPOSE
Contour mode works much better with 1 iteration per update. The RPR2 render engine uses incremental iterations number with max iterations setting hidden from UI.
Make Contour use 1 iteration per view update.

### EFFECT OF CHANGE
- Contour mode iterations per update limited to 1 for better results.

### TECHNICAL STEPS
- stored Contour mode usage flag in render engine;
- disabled iterations number increment for Contour mode.


src/rprblender/engine/render_engine.py
```diff
@@ -145,7 +147,7 @@ def _render(self):
                 break
 
             render_iteration += 1
-            if render_iteration > 1 and self.render_update_samples < MAX_RENDER_ITERATIONS:
+            if render_iteration > 1 and self.render_update_samples < MAX_RENDER_ITERATIONS and not self.use_contour:
```
bsavery, line 150: move self.use_contour is true, other tests are unescessary.  Move first for logic?

bnagirniak (approved):
Works ok, but much more slowly because of rendering 1 iteration. Is this what we want?

bsavery:
> Works ok, but much more slowly because of rendering 1 iteration. Is this what we want?

This is needed for contour mode

PR #148, opened 2020-11-26
[RPRPBLND-1062] Fixed negative Coating Transmission Color values artifacts

ardraeiss:
### PURPOSE
Uber shader node Coating Transmission Color input can receive negative values(as a result of calculations). If Coating Thickness is greater than 0.0 it would result in negative color render artifacts on mesh outline areas, that looks like bright white pixels with (inf, inf, inf, 1.0) values.
If Composition post-processing is used, for example Glare, it could completely ruin the render result.
Fix it for Full and Legacy modes.

### EFFECT OF CHANGE
- fixed render and post-processing artifacts in case of negative Uber shader node Coating Transmission Color input values.

### TECHNICAL STEPS
- added max(0.0, CoatingTransmissionColor) operation to remove negative input values to the Uber node export.

### NOTES FOR REVIEWERS
Use scene attached for testing.
[coating_thickness_with_negative_transmission_color.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/files/5604056/coating_thickness_with_negative_transmission_color.zip)


src/rprblender/nodes/rpr_nodes.py
```diff
@@ -591,13 +591,15 @@ def set_normal(normal_socket_key, use_shader_normal, rprx_input):
                 coating_roughness = self.get_input_value('Coating Roughness')
                 coating_thickness = self.get_input_value('Coating Thickness')
                 coating_transmission_color = self.get_input_value('Coating Transmission Color')
+                # remove transmission color negative values to prevent render artefacts
+                coating_transmission_color_fixed = coating_transmission_color.max(0.0)
```
bsavery, line 595: Change name to coating_transmission_color_clamped

bnagirniak (approved):

PR #149, opened 2020-11-27
[RPRBLND-1435] Hair BSDF shader for RPR2

ardraeiss:
### PURPOSE
RPR2 has no support for Ward shader node used by Hair BSDF Reflection mode.
Recreating Ward using complicated Math nodes graph would hit performance too hard. This PR approximates reflection part of Hair BSDF using anisotropic Reflection of Uber node.

### EFFECT OF CHANGE
- fixed Hair BSDF shader node Reflection mode.

### TECHNICAL STEPS
- added export_rpr2 method to node parser to hande RPR2 specific exports. Reuses regular export by default;
- split HairBSDF nodes creation to separate methods;
- clamped input roughness values in a way Cycles does it to the [0.001..1.0] range;
- used average roughness value;
- used Uber with only anisotropic Reflection enabled to represent Reflection mode of Hair BSDF shader node.

### NOTES FOR REVIEWERS
- Ward node would have been a better approximation of Reflection mode but recreating it would mean too complicated Math nodes usage that would hit performance heavily. This PR
- values used are rough approximations.


bnagirniak (approved):
Looks ok. Also thank you for adding export_rpr2() to NodeParser class.

PR #151, opened 2020-12-01
Use SDK v2.1.8 release

ardraeiss:
### PURPOSE
Newest core SDK version released, 2.1.8.
Use it

### EFFECT OF CHANGE
- updated RPR SDK to the latest version 2.1.8.

### TECHNICAL STEPS
- switched SDK to latest version;
- disabled Python wrap building for RPR_CONTEXT_CREATE_MESH_EDITABLE_FUNC_NAME.

### NOTES FOR REVIEWERS
- The RPR_CONTEXT_CREATE_MESH_EDITABLE_FUNC_NAME definition is an alias for rprContextCreateMeshEditable. Builder fails on it so it was added to "skip list". An additional work would be required to see if it should be wrapped in a different way.


src/bindings/pyrpr/src/pyrprapi.py
```diff
@@ -663,7 +663,8 @@ def get_rif_sdk(base=Path()):
                  'RPR_SHAPE_SET_LIGHTMAP_CHART_INDEX_FUNC_NAME',
                  'RPR_MESH_UPDATE_FUNC_NAME',
                  'RPR_SHAPE_SET_TRANSFORM_BATCH_FUNC_NAME',
-                 'rprDirectionalLightSetRasterShadowSplits']
+                 'rprDirectionalLightSetRasterShadowSplits',
+                 'RPR_CONTEXT_CREATE_MESH_EDITABLE_FUNC_NAME',]
```
bsavery, line 667: Currious.  What does this function do?

ardraeiss: It's from the RadeonProRender_Baikal.h headers file. Other entries in that list are from the same header as well.

Dizzyspb (approved):
Core 2.1.8 integration is tested by QA. Approved for merge.

bnagirniak (approved):

PR #152, opened 2020-12-01
[RPRBLND-1556] export Contour settings to json

ardraeiss:
### PURPOSE
Export Contour settings to RPR json when Export to rpr is used.

### EFFECT OF CHANGE
- added Contour settings to the "Export to rpr". In GPU mode only.

### TECHNICAL STEPS
- added RPR2 exporter for Full mode;
- added Contour settings to RPR file exporter;


bnagirniak (approved):
Didn't test, code is ok.

PR #156, opened 2020-12-07
Rprblnd 1519 contour in viewport

ardraeiss:
### PURPOSE
Enables Contour rendering mode in Viewport.

### EFFECT OF CHANGE
- added Contour mode support to the Viewport render.

### TECHNICAL STEPS
- used Contour mode in Viewport render;
- set Viewport update iterations number to 1 is Contour is used.

### NOTES FOR REVIEWERS
Enabling/disabling Contour with active viewport render would require full scene re-sync and render restart since the whole context should be set in a proper mode before RPR scene is created. Thus it doesn't supported at the moment.


bnagirniak (approved):
Tested - works ok.

PR #159, opened 2020-12-10
Engine version in export json

ardraeiss:
### PURPOSE
Scene exported to RPR file missed the render engine version used for export.
Add it.

### EFFECT OF CHANGE
- added export engine type inf to the RPR file export.

### TECHNICAL STEPS
- added library name used for scene export to the json.


src/rprblender/engine/export_engine.py
```diff
@@ -36,6 +36,11 @@
 
 class ExportEngine(Engine):
     TYPE = 'EXPORT'
+    ENGINE_LIBS = {
```
bnagirniak, line 39: I don't like the idea to store this constant here. I think better just to add code of getting engine_lib_name inside save_json() function.
So, changes should be applied only in save_json(), nothing else has to be touched.

bnagirniak (changes requested):
Looks like easier way would be just add getting engine_lib_name inside save_json(), nothing else has to be touched.

src/rprblender/operators/export_scene.py
```diff
@@ -138,15 +139,26 @@ def execute(self, context):
...
     def export_scene_to_file(self, context, scene, filepath, filepath_json, flags):
-        if scene.rpr.render_quality == 'FULL':
+        if scene.rpr.render_quality == 'FULL':  # Export Legacy mode using RPR1
             exporter = ExportEngine()
-        else:
+            engine_lib_name = {
+                'Windows': "RadeonProRender64.dll",
+                'Darwin': "libRadeonProRender64.dylib",
+                'Linux': "libRadeonProRender64.so",
+            }.get(OS, '')
```
bnagirniak, line 148: Why use get() instead direct {...}[OS] ?

bnagirniak (commented):
Small code improvement could be applied.

bnagirniak (approved):

PR #160, opened 2020-12-11
RPR 2 restarts to viewport

bsavery:
### PURPOSE
With RPR2, viewport would restart with any change of window (including switching properties window tabs).  This is "unacceptable behavior" according to some.  Certainly it is undesired.  

Also the "stop render" time was too long when switching out of viewport render mode

### EFFECT OF CHANGE
Viewport no longer restarts when changing tabs in properties

Also the time to stop the viewport render mode is much shorter

### NOTES FOR REVIEWERS
@bnagirniak  I'm not sure why this code was here before. Please correct me if this is wrong.  (the part that the render is restarted even if no updates)
Also there is a bit of a lag on my laptop when switching properties tabs if in the middle of 32 samples.  If there's a way to address this please let me know.


src/rprblender/engine/viewport_engine_2.py
```diff
@@ -43,12 +43,13 @@ def __init__(self, rpr_engine):
...
     def stop_render(self):
-        self.rpr_context.set_render_update_callback(None)
-
+        self.restart_render_event.set()
+        
         super().stop_render()
 
         self.resolve_event.set()
         self.resolve_thread.join()
+        self.rpr_context.set_render_update_callback(None)
```
bnagirniak, line 52: It raises error here: `AttributeError: 'NoneType' object has no attribute 'set_render_update_callback'` because super().stop_render() sets self.rpr_context = None.

bnagirniak: This function can be written in this way
```
    def stop_render(self):
        self.is_finished = True
        self.restart_render_event.set()
        self.resolve_event.set()
        self.sync_render_thread.join()
        self.resolve_thread.join()

        self.rpr_context.set_render_update_callback(None)
        self.rpr_context = None
        self.image_filter = None

```

bnagirniak (changes requested):
Found error and added suggestion to fix.

bnagirniak (approved):

ardraeiss (approved):
Code is fine.

The old behavior might be undesirable to some users, but the new one hangs for a few seconds on any attempt to switch tabs while viewport render isn't finished. Try on BMV Blender demo scene.

bsavery:
> The old behavior might be undesirable to some users, but the new one hangs for a few seconds on any attempt to switch tabs while viewport render isn't finished. Try on BMV Blender demo scene.

Yes this is where I was hoping for some ideas.  @bnagirniak any ideas?

bnagirniak:
> Yes this is where I was hoping for some ideas. @bnagirniak any ideas?

I'll look and think about it tomorrow.

bnagirniak:
Found and fixed issue with switching tabs. Lag was in `with self.render_lock:` which was called even if no updates are coming.

ardraeiss (approved):
It lags no more, great

PR #165, opened 2020-12-13
Cryptomatte AOV enable

bsavery:
### PURPOSE
To enable Cryptomatte AOVs, a powerful way of creating and selecting certain parts of the render in compositing.  
Note.  I followed this tutorial of usage in blender compositor:  https://www.youtube.com/watch?v=VBJQM-gIob8&ab_channel=FlippedNormals

### EFFECT OF CHANGE
Two AOV options are added "Material" and Object Cryptomatte AOVs.  Each creates 3 aovs that can be hooked up to the cryptomatte node in the compositor and then allowed to pick whatever objects or material you want

### NOTES FOR REVIEWERS
I had to make a separate aovs_info for cryptomatte.  Why?  Because BoolVector properties can only go to 32 in blender. ;).   With the 6 needed for cryptomatte we went over this limit.  I suppose I could hack it so that 1 item in the aovs_info list could create more than one aov, but that feels hacky.

### Known issues:
The aovs are black on macOS with GPU rendering, ok for CPU.


bsavery:
Screen Shot 2020-12-12 at 5 08 14 PM
Using this with BMW scene and selecting the wheel

bnagirniak:
Should we have cryptomate panel in UI or something like this? Where in UI cryptomate should be applied?

bsavery:
> Should we have cryptomate panel in UI or something like this? Where in UI cryptomate should be applied?

I just added aov buttons for it.  We could add a separate panel but I don't think it's necessary.

bnagirniak:
Where are such buttons? Can't find them in view layer properties tab, also can't find such code in code changes.

bsavery:
Fixed.

bnagirniak (approved):
Tested - it works now. Would be good also to hear some overview from QA team.

ardraeiss (approved):
Code is fine

bsavery:
> Tested - it works now. Would be good also to hear some overview from QA team.

Will ping them

ARozghon:
was checked manually and it's work as expected if compare results with Cycles

PR #166, opened 2020-12-13
Add "Use image cache" flag for export

bsavery:
### PURPOSE
Latest RPR core has an option for exporting images using the image cache.  Add to export options

### EFFECT OF CHANGE
Users can now export cached images.  Users should note that the version of RPRSRender.exe MUST match the version in the plugin


bnagirniak (approved):
Tested - works ok.

ardraeiss (approved):
Fine.

PR #169, opened 2020-12-16
[RPRBLND-1570] Disabled Contour mode in CPU and combo modes

ardraeiss:
### PURPOSE
Contour mode works correctly in GPU mode only. Disable it if CPU selected as as single or one of render devices.

### EFFECT OF CHANGE
- disabled Contour rendering if CPU selected as one of render devices.

### TECHNICAL STEPS
- disabled Contour rendering if CPU is enabled as a render device;
- used Final mode settings for RPR scene export.


bnagirniak (approved):
Tested - works ok.

bsavery:
This is fine.  I'm trying to get some clarity from the RPR team when CPU would be supported.  You know me I hate "workarounds"

PR #171, opened 2020-12-18
Fixed UI descriptions

ardraeiss:
### PURPOSE
Fixed tiled render description.

### EFFECT OF CHANGE
- fixed tiled render description to display it's available in Legacy mode only;
- removed double trailing dots from field descriptions.

### TECHNICAL STEPS
- changed tile render description to display it's available in Legacy mode only;
- removed trailing dots from descriptions.


bnagirniak (approved):
Useful changes, thanks.

PR #176, opened 2021-01-05
Box mapping on Image Texture node

bsavery:
### PURPOSE
Enable "Box mapping" option in the image texture node

### EFFECT OF CHANGE
Box mapping now works in the image texture node

### TECHNICAL STEPS
I followed common triplanar mapping code, ex: https://catlikecoding.com/unity/tutorials/advanced-rendering/triplanar-mapping/

Could not get the RPR triplanar node to work like I wanted.  

### NOTES FOR REVIEWERS
Normally, you use the "Generated" texture coordinate tied into the image texture for this.  I think the Generated coord only works in RPR 1 right now (legacy mode).  P_Local lookup doesn't work in RPR 2?


bsavery:
Cycles:
cycles
RPR:
rpr

src/rprblender/nodes/blender_nodes.py
```diff
@@ -666,19 +666,59 @@ def _export_image_node(self, wrap_mapping):
...
+            tex_xz = self.create_node(pyrpr.MATERIAL_NODE_IMAGE_TEXTURE, {
+                pyrpr.MATERIAL_INPUT_DATA: rpr_image,
+                pyrpr.MATERIAL_INPUT_UV: p_xz
+            })
 
-        vector = self.get_input_link('Vector')
-        if vector:
-            rpr_node.set_input(pyrpr.MATERIAL_INPUT_UV, vector)
+            # calculate blend factor
+            blend = 1.0 / self.node.projection_blend if not math.isclose(self.node.projection_blend, 0.0) else 999.9
```
bnagirniak, line 698: blend value is never used there, can be removed

bnagirniak (changes requested):
Tested and found inconsistency of texture mapping between RPR and cycles:
RPR:
image
Cycles:
image

Also found defined value that is never used.

src/rprblender/nodes/blender_nodes.py
bsavery: ah this is an error, should be multiplied by weights.

bnagirniak (approved):
I fixed inconsistency issue with Cycles, please review.

bsavery:
> This looks closer to the Cycles(if Blend is set to 1.0 for Cycles).

What about other blend values?

PR #178, opened 2021-01-08
Add ability to rotate environment overrides independently

bsavery:
### PURPOSE
Mike P requests ability to rotate environment overrides independently

### EFFECT OF CHANGE
Each override has its own rotation and can be rotated separately.  If not enabled, use the main rotation


src/rprblender/export/world.py
```diff
@@ -247,6 +248,8 @@ def set_override(override_type):
                 override_data.image = image.name
             else:
                 override_data.color = tuple(color)
+            override_data.rotation = getattr(rpr, f'{override_type}_rotation') \
```
bnagirniak, line 251: Convert it to tuple to make it comparable in BackplateData, should be:
```
override_data.rotation = tuple(getattr(rpr, f'{override_type}_rotation')) \
....
```
It fixes interactive changing rotation in viewport render.

bnagirniak (changes requested):
Found how to fix interactive override rotation in viewport render.

ardraeiss (approved):
Now it's good.

PR #181, opened 2021-01-13
Ocio rebase

bsavery:
PURPOSE
Blender includes a config.ocio file for texture color spaces. We can use this to define the color space for textures and set the ocio colorspace for image textures for RPR2

EFFECT OF CHANGE
Texture Colorspace is respected better in RPR2, particularly for non-srgb or raw color spaces

NOTES FOR REVIEWERS
A further change could be to use the OCIO render space to use the Filmic setting, not sure if this would help us any.
Also Blender actually allows an environment variable to do a custom config.ocio file. That's not supported but could be.

This include the MacOS compilation fix too..


bnagirniak (commented):
Tested - works ok.

bnagirniak (approved):

PR #183, opened 2021-01-14
Noise seed for animation renders

bsavery:
### PURPOSE
When rendering an animation, the noise sampling pattern is the same for each frame, giving an unpleasant "swimming" texture if the camera is moving
 
### EFFECT OF CHANGE
For animation renders, noise sampling pattern is not the same in each frame. 

### TECHNICAL STEPS
To do this we just save the CONTEXT_FRAMECOUNT setting between frames.  The issue was because they started from 0 each frame.


bnagirniak (approved):
Looks ok, also have good feedback from QA.

ardraeiss (approved):

PR #187, opened 2021-01-18
RPRBLND-1583: Viewport optimizations for Northstar

bnagirniak:
### PURPOSE
We should try and improve the viewport rendering speed in Northstar:
1.  Enable viewport scaling
2.  Only resolve color AOV for intermediate iterations. Other AOVs should be resolved only at last iteration for image filter needs.

### EFFECT OF CHANGE
Applied adaptive viewport resolution for Full render quality. Applied internal optimizations for viewport rendering.
Enabled Adapt Viewport Resoltion option in Viewport & Preview Sampling panel.

### TECHNICAL STEPS
1. Moved adaptive resolution code in ViewportEngine to _adapt_resize() function, made it available to use in ViewportEngine2.
2. Overrided _resize() in ViewportEngine2.
3. Implemented algorithm of adaptive resolution in ViewportEngine2._do_render().
4. Enabled UI Adapt Viewport Resoltion option.
5. Improved RPRContext.resolve() with aovs parameter. Optimized resolving at intermediate iterations.

### NOTES FOR REVIEWERS
I found that render time at second iteration with CONTEXT_PREVIEW=3 is approximately equal to render time per iteration with CONTEXT_PREVIEW=0 and CONTEXT_ITERATIONS=32. This is quite rough approximation and depends by the scene, but I used this render time as estimation for calculating requested_adapt_ratio value.

### DIFF
--- a/src/rprblender/engine/context.py
+++ b/src/rprblender/engine/context.py
@@ -149,11 +149,19 @@ class RPRContext:
 
         return self.frame_buffers_aovs[pyrpr.AOV_COLOR]['res']
 
-    def resolve(self):
-        for aov, fbs in self.frame_buffers_aovs.items():
-            fbs['aov'].resolve(fbs['res'], aov != pyrpr.AOV_SHADOW_CATCHER)
+    def resolve(self, aovs=None):
+        if aovs:
+            for aov in aovs:
+                fbs = self.frame_buffers_aovs[aov]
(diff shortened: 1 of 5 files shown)


ardraeiss (approved):
OK

bsavery:
Looks great.  Performance is much better!  One question.  Should we resolve what the viewing AOV is not nescessarily color?

PR #188, opened 2021-01-21
Enabled texture compression for Full render quality mode

ardraeiss:
### PURPOSE
Display and use texture compression option for Full render quality mode.

### EFFECT OF CHANGE
- added "Use texture compression" option to the Render Quality settings tab in Full mode.

### TECHNICAL STEPS
- added texture compression option;
- set texture compression by user settings in FULL2 render quality mode.


bnagirniak (approved):
Code is ok.

bsavery:
My question is if we want to move this to somewhere more hidden.  Ie. most users should not use this.  We could put it by the texture cache setting.

ardraeiss:
The SDK fix is required for this feature to have effect on render results. Closing this PR until that moment.

ardraeiss:
> 
> 
> My question is if we want to move this to somewhere more hidden. Ie. most users should not use this. We could put it by the texture cache setting.

Sure, I'll move it to the Debug panel.

PR #190, opened 2021-01-26
[RPRBLND-1625] visibility of linked collection objects

ardraeiss:
### PURPOSE
Linked collection object instance applied incorrect visibility settings(from source object parent).

### EFFECT OF CHANGE
- fixed instance objects incorrect ray visibility settings usage.

### TECHNICAL STEPS
- applied source object visibility settings instead of the parent settings.


bnagirniak (approved):
Looks ok

PR #192, opened 2021-01-27
[RPRBLND-1643] Fixed enabling disabled instances

ardraeiss:
### PURPOSE
Instances are not updated correctly in viewport on disabling-enabling - it correctly removed when disabled but isn't displayed on enabling.
There is also an issue with material override application/removal: it resets the object rays visibility settings and isn't updated on instances.

### EFFECT OF CHANGE
- fixed material override resetting object visibility settings;
- fixed instances enabling in viewport.

### TECHNICAL STEPS
- called for instance sync for visible instance that were not present before on scene update;
- simplified objects and instances update on scene update;
- removed visibility settings reset on scene update.


src/rprblender/engine/viewport_engine.py
```diff
@@ -868,7 +868,11 @@ def sync_objects_collection(self, depsgraph):
             self.sync_collection_objects(depsgraph, unchanged_meshes_keys,
                                          material_override)
 
-            self.sync_collection_instances(depsgraph, unchanged_meshes_keys,
+            rpr_instance_keys = set(key for key, obj in self.rpr_context.objects.items()
```
bsavery, line 871: Would it make more sense for rpr_context to keep a list of instances ?

bnagirniak (approved):
Tested - works ok.

PR #193, opened 2021-01-28
Added Mipmap LOD offset settings

ardraeiss:
### PURPOSE
The "Full" mode supports option of changing mipmap level of details switching distance for fine tuning quality-vs-memory.
Use it.

### EFFECT OF CHANGE
- added option of changing images mipmap level of detail switching distance to the Debug panel, for fine tuning of image quality vs memory consumption. The bigger the value the sharper textures and bigger memory usage will be. "Full" render quality mode only.

### TECHNICAL STEPS
- added float LOD offset property;
- displayed property in FULL2 mode only, in Debug render settings panel;
- set context property RPR_CONTEXT_MIPMAP_LOD_OFFSET to the offset property, in FULL2 mode only.


ardraeiss:
Closing since this feature is to be discussed.

PR #201, opened 2021-02-04
[RPRBLND-1465] RIF Transparent background for Full render quality

ardraeiss:
### PURPOSE
Full quality mode(RPR2) doesn't support composite operations thus it doesn't have a transparent background support.
Use RIF to filter out the background for RPR2.

### EFFECT OF CHANGE
- added basic Transparent Background support for "Full" render quality.

### TECHNICAL STEPS
- added custom RIF filter to filter out the background from color image;
- added background_filter to the render engine. Set it up with custom transparent background RIF filter for Full quality mode only;
- applied background_filter on rendered image if filter is present in final mode.

### NOTES FOR REVIEWERS
- This version doesn't work with RPR2 shadow catcher. This will be added as a separate PR in the next task.
- The viewport doesn't truly applies transparent background in Legacy mode so it was left for the next filters and denoisers updates.


bnagirniak (approved):
Code is ok, Tested - works ok for final render for Full render quality. For viewport it is not implemented.

PR #204, opened 2021-02-09
[RPRBLND-1657] Disable RIF Transparent Background on OSX

ardraeiss:
### PURPOSE
New Full render quality Transparent Background causes Blender crash on OSX. Disable it for now until filter is fixed for OSX.

### EFFECT OF CHANGE
- disabled Transparent Background on OSX in non-Legacy render qualities;
- enabled Transparent Background on High, Medium and Low render qualities.

### TECHNICAL STEPS
- hidden Transparent background panel on OSX for non-Legacy render qualities;
- disabled RIF transparent background filter on OSX;
- enabled Transparent background for Hybrid modes.


bsavery:
What's the issue with getting this to work on osx?

ardraeiss:
Will use as a fallback solution in case if proper fix is delayed.

PR #207, opened 2021-02-16
[RPRBLND-1657] Fixed RIF Transparent Background filter on OSX

ardraeiss:
### PURPOSE
OSX handles the RIF filter output buffer differently than Windows or Linux causing SEG_FAULT crash on transparent background filter application.
This PR fixes this behavior by switching filter to default input and output buffer names(“inputImage” and “outputImage”) instead of ImageFilterSetParameterImage usage.

### EFFECT OF CHANGE
- fixed Transparent Background crash in Full render quality;
- enabled Transparent Background for High, Medium and Low render qualities.

### TECHNICAL STEPS
- renamed input and output RIF filter buffers so Metal will handle filter application correctly;
- enabled RIF-based transparent background filter for Hybrid modes;
- added support for more useful argument types - 2-3-4-length tuples of unsigned integers or floats.

### NOTES
- RIF documentation for USER_DEFINED filter type should be adjusted to use correct input and output buffer names.


src/bindings/pyrpr/src/pyrprimagefilters.py
```diff
@@ -305,9 +305,25 @@ def set_parameter(self, name, value):
         elif isinstance(value, Image):
             ImageFilterSetParameterImage(self, pyrpr.encode(name), value)
             self.parameters[name] = value
-        elif isinstance(value, tuple) and len(value) == 2:
+        elif isinstance(value, tuple) and len(value) > 0:
```
bsavery, line 308: please call len(value) once only.

src/rprblender/engine/render_engine.py
```diff
@@ -476,7 +476,7 @@ def sync(self, depsgraph):
         self.setup_image_filter(image_filter_settings)
 
         # Shadow catcher
-        if scene.rpr.render_quality == 'FULL2':
+        if scene.rpr.render_quality != 'FULL':
```
bsavery, line 479: so this should work for hybrid too?

ardraeiss: it uses RIF only so it works on Hybrid

src/rprblender/engine/image_filter.py
```diff
@@ -112,28 +112,26 @@ def get_data(self):
...
+    def setup_alpha_filter(self, alpha):
         """ Apply transparent background by setting output image alpha by alpha value """
         result = self.context.create_filter(rif.IMAGE_FILTER_USER_DEFINED)
 
         # redefine image alpha channel by alpha value
         code = \
             "int2 coord;" \
-            "GET_COORD_OR_RETURN(coord, GET_BUFFER_SIZE(outBuf));" \
-            "vec4 pixel = ReadPixelTyped(inputBuf, coord.x, coord.y);" \
+            "GET_COORD_OR_RETURN(coord, GET_BUFFER_SIZE(outputImage));" \
```
bnagirniak, line 122: Looks like using 3 quotes syntax for multiline string would suit better here:
code = """
.....
"""

bnagirniak (commented):
Left comment about code improvement.

src/rprblender/engine/image_filter.py
bsavery: +1

ardraeiss: Yeah, that looks much better

bnagirniak (approved):
Code is ok.

PR #208, opened 2021-02-19
RPRBLND-1634: Light power animation not update in realtime.

bnagirniak:
### PURPOSE
Lights power animation shows only the first frame of animation when viewport render start.

### EFFECT OF CHANGE
Fixed light update during viewport rendering with animated light settings like power, color, etc).

### TECHNICAL STEPS
Fixed, by adding bpy.types.Light to sync_update() checking.


ardraeiss (approved):
Great

PR #215, opened 2021-03-05
RPRBLND-1664: Add support for ML denoiser + upscale

bnagirniak:
### PURPOSE
RIF has upscale ML filter, which provides us to render at lower resolution in viewport and upscaling rendered image.

### EFFECT OF CHANGE
Enabled upscaling in viewport render, which provides faster render at 2-times lower resolution and upscaling rendered image. Added correspondent option in viewport sampling tab.

### TECHNICAL STEPS
1. Added class ImageFilterUpscale.
2. Added Engine.upscale_filter and correspondent methods.
3. Added viewport_resolution_upscale property in user settings, added this property in UI in viewport sampling tab.
4. Implemented upscale_filter usage in ViewportEngine and ViewportEngine2.

### DIFF
--- a/src/rprblender/ui/render.py
+++ b/src/rprblender/ui/render.py
@@ -170,6 +170,8 @@ class RPR_RENDER_PT_viewport_limits(RPR_Panel):
 
         col.prop(settings, 'use_gl_interop')
 
+        col.prop(settings, 'viewport_resolution_upscale')
+
         col.separator()
         col.prop(limits, 'preview_samples')
         col.prop(limits, 'preview_update_samples')
--- a/src/rprblender/ui/render.py
+++ b/src/rprblender/ui/render.py
@@ -170,7 +170,7 @@ class RPR_RENDER_PT_viewport_limits(RPR_Panel):
 
         col.prop(settings, 'use_gl_interop')
 
-        col.prop(settings, 'viewport_resolution_upscale')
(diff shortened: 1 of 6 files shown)


bsavery:
Looks ok I want to review performance and UI next week.

src/rprblender/ui/render.py
```diff
@@ -170,6 +170,8 @@ def draw(self, context):
 
         col.prop(settings, 'use_gl_interop')
 
+        col.prop(settings, 'viewport_resolution_upscale')
```
ardraeiss, line 173: Maybe it should be hidden or disabled for Legacy and Hybrid modes?

ardraeiss (approved):
It's OK, maybe some UI adjustment could be done since this is Full mode only feature.

src/rprblender/ui/render.py
bnagirniak: It is working in all modes.

bsavery (changes requested):
I'm going to propose a simple UI change.  Make a button in viewport settings

"Viewport Denoising and upsampling" Which enables both in the viewport.  enabled by default

bnagirniak:
Made one option for viewport denosing and upscaling with ML denosier.

ardraeiss (approved):

PR #218, opened 2021-03-10
Improved helper library loading error message

ardraeiss:
### PURPOSE
It's unclear from Blender error message what addon version has issue with helper library loading. This PR expands the OSError traceback with addon, core and RIF version numbers and library path so it would be easier to figure out loading issue reasong from exception window screenshot.

### EFFECT OF CHANGE
- added addon, RPR and RIF version numbers and the addon helper library path to the library loading exception.

### TECHNICAL STEPS
- raised Exception on Helper lib OSError loading exception with versions info;
- attached it to the original exception traceback.

### NOTES
The result will look like this
image
Maybe it's too big.


src/rprblender/utils/helper_lib.py
```diff
@@ -60,7 +61,15 @@ def init():
 
     lib_path = next(p / lib_name for p in paths if (p / lib_name).is_file())
     log('Load lib', lib_path)
-    lib = ctypes.cdll.LoadLibrary(str(lib_path))
+    try:
+        lib = ctypes.cdll.LoadLibrary(str(lib_path))
+    except OSError as e:  # expand the traceback info with the exact addon version and library name
+        raise Exception(f"Failed to load library {lib_path}",
```
bnagirniak, line 67: What do you think of just logging this error and provide plugin to be in working state? Of course openvdb and sun&sky functionality won't be working, but everything else will be working. 
@bsavery what do you think about it?

bnagirniak (commented):
Added improvement comment question.

src/rprblender/utils/helper_lib.py
bsavery: I mean, it depends on the library.  If RPR lib fails that's bad.  But hard to tell if other libraries don't fail what is going to work and what won't

bsavery: So a "borked" load is hard to predict.  and might be ok just to fail

bnagirniak (approved):

PR #219, opened 2021-03-10
Use Cryptomatte AOVs in Full mode only

ardraeiss:
### PURPOSE
Cryptomatte AOVs are fully supported by Full render quality mode at the moment. The Legacy render doesn't work if they are enabled.
With this PR they'll be actually used only in Full render quality mode.

### EFFECT OF CHANGE
- fixed: enabling Cryptomatte AOVs no longer blocks Legacy mode render.

### TECHNICAL STEPS
- disabled Cryptomatte AOVs activation for all modes other than Full.


bnagirniak (approved):
Code looks ok

PR #223, opened 2021-03-15
Rprblnd 1669 Enable arithmetic operations in Hybrid render modes

ardraeiss:
### PURPOSE
Enables usage of arithmetic operations currently supported by Hybrid.

### EFFECT OF CHANGE
- added support for simple Math/MixRGB/Vector Math nodes operations in High, Medium and Low render qualities.

### TECHNICAL STEPS
- enabled arithmetic operations usage in Hybrid modes;
- explicitly ignored unsupported arithmetics for Hybrid;
- updated Hair BSDF node roughness calculations for Hybrid;
- cleaned up some typos, code style warnings and unused imports.

### NOTES
- Hybrid uses regular export method if export_hybrid isn't present thus all regular arithmetic users will use them in Hybrid as well unless explicitly set to different export usage.
- Hybrid arithmetic results are different for AMD and NVidia cards;
- performance isn't that good for the main GPU used by OS.


bsavery (changes requested):
I'm unclear how this works.  Shouldn't many nodes need the export_hybrid methods edited?

bnagirniak (approved):
Looks ok.

bsavery (approved):

PR #225, opened 2021-03-16
[RPRBLND-1669] Use UV Lookup in Hybrid

ardraeiss:
### PURPOSE
With latest core changes the UV mode of Lookup node is working in Hybrid. Use it where it's applicable.

### EFFECT OF CHANGE
- added support for Mapping node in High, Medium and Low render qualities;
- added support for Texture Coordinates UV mode in in High, Medium and Low render qualities;

### TECHNICAL STEPS
- enabled Mapping node export in Hybrid;
- checked the Texture Coordinates UV mode works correctly;
- added RPR LookUp node UV and UV1 modes support for Hybrid, both as just UV.


bnagirniak (approved):
Code is ok.

PR #234, opened 2021-03-24
[RPRBLND-1685] fixed RIF transparent background with tiles

ardraeiss:
### PURPOSE
Fixed incorrect result of Full mode Transparent Background with Tiled render mode.

### EFFECT OF CHANGE
- fixed tiled render with Transparent Background in Full render quality.

### TECHNICAL STEPS
- cut out the correct filtered image area for full-sized render and for each each tile;
- called for additional filter application after denoising.


bnagirniak (approved):
Tested - works ok.

PR #246, opened 2021-04-02
README: Remove epydoc recommendation

ferdnyc:
### PURPOSE
Update the repo README to no longer recommend epydoc (which isn't even Python 3 compatible).

### EFFECT OF CHANGE
README no longer directs developers to outdated software that can't actually be used.


bsavery:
Thanks @ferdnyc

PR #262, opened 2021-04-21
Add support for setting texture compression.  Default to false

bsavery:
### PURPOSE
Many issues persist with texture compression with RPR.  Setting to false fixes many

### EFFECT OF CHANGE
Less issues with texture compression

### TECHNICAL STEPS
we set this globally rather than per texture.  Thus is set on rpr_context.

### DIFF
--- a/src/rprblender/engine/context.py
+++ b/src/rprblender/engine/context.py
@@ -80,6 +80,9 @@ class RPRContext:
         self.use_reflection_catcher = False
         self.use_transparent_background = False
 
+        # texture compression used when images created
+        self.texture_compression = False
+
     def init(self, context_flags, context_props, use_contour_integrator=False):
         self.context = self._Context(context_flags, context_props)
         self.material_system = pyrpr.MaterialSystem(self.context)
@@ -452,12 +455,14 @@ class RPRContext:
 
     def create_image_file(self, key, filepath):
         image = pyrpr.ImageFile(self.context, filepath)
+        image.set_compression = self.compress_textures
         if key:
             self.images[key] = image
         return image
 
     def create_image_data(self, key, data):
         image = pyrpr.ImageData(self.context, data)
+        image.set_compression = self.compress_textures
         if key:
             self.images[key] = image
         return image
--- a/src/rprblender/engine/viewport_engine.py
+++ b/src/rprblender/engine/viewport_engine.py
@@ -516,6 +516,7 @@ class ViewportEngine(Engine):
         self.rpr_context.set_parameter(pyrpr.CONTEXT_ITERATIONS, 1)
         scene.rpr.export_render_mode(self.rpr_context)
         scene.rpr.export_ray_depth(self.rpr_context)
+        self.rpr_context.texture_compression = scene.rpr.texture_compression
         scene.rpr.export_pixel_filter(self.rpr_context)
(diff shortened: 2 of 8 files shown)


src/rprblender/engine/viewport_engine.py
```diff
@@ -516,6 +516,7 @@ def sync(self, context, depsgraph):
         self.rpr_context.set_parameter(pyrpr.CONTEXT_ITERATIONS, 1)
         scene.rpr.export_render_mode(self.rpr_context)
         scene.rpr.export_ray_depth(self.rpr_context)
+        self.rpr_context.texture_compression = scene.rpr.texture_compression
```
ardraeiss, line 519: Wouldn't it be better to set the compression option inside the scene.rpr.init_rpr_context call?

ardraeiss (approved):
Looks OK

src/rprblender/engine/viewport_engine.py
bsavery: this setting could change after the context is created

bsavery:
@bnagirniak please review.

src/rprblender/engine/context.py
```diff
@@ -452,12 +455,14 @@ def set_material_node_as_material(self, key, material_node):
 
     def create_image_file(self, key, filepath):
         image = pyrpr.ImageFile(self.context, filepath)
+        image.set_compression = self.compress_textures
```
bnagirniak, line 458: set_compression is a method in Image class. Seems must be image.set_compression(self.compress_textures)

src/rprblender/engine/context.py
```diff
@@ -452,12 +455,14 @@ def set_material_node_as_material(self, key, material_node):
...
         if key:
             self.images[key] = image
         return image
 
     def create_image_data(self, key, data):
         image = pyrpr.ImageData(self.context, data)
+        image.set_compression = self.compress_textures
```
bnagirniak, line 465: image.set_compression(self.compress_textures) ?

bnagirniak (changes requested):
Looks like there are errors in code

src/rprblender/engine/context.py
ardraeiss, line 458: Also AttributeError:
AttributeError: 'RPRContext2' object has no attribute 'compress_textures'

ardraeiss (changes requested):
The self.con

src/rprblender/ui/render.py
```diff
@@ -196,6 +196,9 @@ def draw(self, context):
         if rpr.render_quality in ('LOW', 'MEDIUM', 'HIGH'):
             self.layout.prop(rpr, 'hybrid_low_mem')
 
+        if rpr.render_quality == 'FULL2':
+            self.layout.prop(rpr, texture_compression)
```
bnagirniak, line 200: use texture_compression in quotas '':
self.layout.prop(rpr, 'texture_compression')

bnagirniak (changes requested):
one more bug found

bnagirniak:
Found unsupported exceptions issues with hybrid modes and decided to fix them.

bnagirniak (approved):
Tested - works ok.

PR #264, opened 2021-05-05
RPRBLND-1711: Support Blender 2.93

bnagirniak:
### PURPOSE
We would like to support Blender 2.93. Blender decided to switch to Python 3.9 + OpenVDB 8.0.1 with 2.93. Therefore we have to build bindings to python 3.9. 

### EFFECT OF CHANGE
Made plugin to work in Blender 2.93+. Known issue: OpenVDB volumes are not supported temporary in 2.93+

### TECHNICAL STEPS
1. Adjusted build process to create bindings for python 3.7 and python 3.9.
2. Made RPRBlenderHelper to be in 2 variants: with OpenVDB support and without OpenVDB support. Adjusted helper_lib to load correct RPRBlenderHelper depending of OpenVDB support.
3. Fixed rpath for libRadeonML in MacOS.

### NOTES FOR REVIEWERS
Blender 2.93 uses OpenVDB 8.0.1 instead 7.0.0 in previous versions. RPRBlenderHelper uses OpenVDB 7.0.0 which prevents us to load it in Blender 2.93. Therefore I decided to separate RPRBlenderHelper into 2 variants: RPRBlenderHelper_vdb.dll (with OpenVDB support) and RPRBlenderHelper.dll (without OpenVDB support). 
I think making OpenVDB support in Blender 2.93 is a separate task. Probably we have to improve Blender API to get volume data directly.


ardraeiss (approved):
OK, waiting for automated build system checks and/or fixes.

build_osx.sh
```diff
@@ -17,6 +17,7 @@ if [ -f "$cxml" ]; then
 	python3.7 src/bindings/pyrpr/src/pyrprapi.py $cxml
 	if [ -f "./bindings-ok" ]; then
 		python3.7 build.py
+		python3.9 build.py
```
bsavery, line 20: this implies that we have to have both?  Can we build just one or the other?

bnagirniak: Yes, we need both currently. I think whole build process isn't optimized and requires to refactor and improve. I suggest to do this as separate task in future.

PR #271, opened 2021-05-18
Refactor Contour rendering functionality and UI RPRBLND-1708

bsavery:
### PURPOSE
Contour Rendering is a bit disjointed in the current usage.  The workflow for users is strange.

### EFFECT OF CHANGE
Now contour rendering (Now called Outline) is moved to the view layer AOV section.  Enable Outline rendering as you would other AOVs.  Your render will take take two passes per view layer, the second doing the outline render.

### TECHNICAL STEPS
There are lots of changes here (and many deletions too!) so I'll try and summarize:
* Moving Contour settings to view layer properties
* Generally removing the setup of contour settings at context creation.  Instead always set contour properties and switch the context integrator when needed
* Setup the render engine to do two passes if contour render is enabled
* Deal with AOVs for just the contour rendering pass if in that mode

### NOTES FOR REVIEWERS
I think as a further TODO I might tackle enabling AOVs in the viewport


bnagirniak (changes requested):
Didn't test, but Artur found following error in this PR
image_2021_05_18T14_39_03_898Z

ardraeiss:
The outline is now looking fine, but the color image is getting completely re-rendered several times instead of only one.

bsavery:
> The outline is now looking fine, but the color image is getting completely re-rendered several times instead of only one.

Fixed.

ardraeiss (changes requested):
Final render looks OK, the Viewport is broken:
2021-05-20 23:36:43,524 ERROR rpr.init [4900]:  'ViewLayerSettings' object has no attribute 'contour' EXCEPTION: Traceback (most recent call last):
  File "D:\Work\RPR\blender_review\RadeonProRenderBlenderAddon\src\rprblender\__init__.py", line 160, in view_update
    self.engine.sync(context, depsgraph)
  File "D:\Work\RPR\blender_review\RadeonProRenderBlenderAddon\src\rprblender\engine\viewport_engine_2.py", line 297, in sync
    super().sync(context, depsgraph)
  File "D:\Work\RPR\blender_review\RadeonProRenderBlenderAddon\src\rprblender\engine\viewport_engine.py", line 494, in sync
    self.view_layer_data.contour.export_contour_settings(self.rpr_context)
AttributeError: 'ViewLayerSettings' object has no attribute 'contour'

ardraeiss:
Fixed Legacy render mode.
Fixed Viewport render, it renders Contour instead of color if Outline enabled in Full mode.
Switching Outline doesn't change active Viewport settings, it would be the TODO.

bsavery:
> Fixed Viewport render, it renders Contour instead of Color if Outline is enabled in Full mode.

I don't think we want this.  

I noted in the TODO that a nice follow on PR would be to make a change to select the AOV (or outline mode) in the viewport options to choose a aov to show.

ardraeiss (approved):

bnagirniak (commented):
Works ok.
I just see a big difference in speed of contour rendering between this branch and develop branch. Contour rendering is much more slowly here. Is it expected?

bsavery:
> Works ok.
> I just see a big difference in speed of contour rendering between this branch and develop branch. Contour rendering is much more slowly here. Is it expected?

No and I'm not sure why this is the case.  Is samples different?  Maybe @ardraeiss can help here since he wrote the initial

bnagirniak (approved):
Made optimization for contour rendering and made some code improvements.

PR #273, opened 2021-05-18
RPRBLND-1668 Deformation blur

bsavery:
### PURPOSE
Add support for Deformation motion blur.  This also refactors motion blur computation

### EFFECT OF CHANGE
Support for deforming objects to be blurred

### TECHNICAL STEPS
I took this this opportunity to refactor the motion blur data collection.  The reason for this was because deformation data needs to be known at mesh creation time.  But we were only caching motion data after mesh creation.  

Now we cache deformation and transformation before creating mesh.  They are cached on the rpr_context, and when creating mesh can get the cached deformation data.  Data is popped from memory when done.  

TODOs here:
Handle actual caching of vertices and exporting for deformation.  @bnagirniak 
Also it seemed we were caching from the previous frame not the next one before?  Strange.


bnagirniak:
Finished implementation, please review.

bnagirniak:
### TECHNICAL STEPS
1. Refactored motion blur data collection: made caching motion blur data and use it during mesh creations, added cache_blur_data() methods in export modules.
2. Added "Deformation blur" option (default false) as addition to "motion blur" option in mesh settings. This is added for optimizing export process.

bnagirniak (approved):
Tested from my side - works ok.

src/rprblender/engine/context.py
```diff
@@ -127,6 +131,9 @@ def clear_scene(self):
 
         self.images = {}
 
+        self.motion_cache = {}
```
bsavery, line 134: if this no longer has deformation, it should be named transform_cache or matrix_cache

src/rprblender/export/mesh.py
```diff
@@ -315,12 +315,26 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, **kwargs):
...
+            data.num_face_vertices,
+            {pyrpr.MESH_MOTION_DIMENSION: 2}
+        )
+    else:
+        rpr_shape = rpr_context.create_mesh(
+            obj_key,
+            data.vertices, data.normals, data.uvs,
+            data.vertex_indices, data.normal_indices, data.uv_indices,
+            data.num_face_vertices
+        )
```
ardraeiss, line 336: Should't there be the None passed as a mesh_info value?
The other way would be to add a default "None" value to it in the rpr_context.create_mesh method header.

bnagirniak: None is not needed here, it is ok here.

ardraeiss (approved):
Looks OK

PR #275, opened 2021-05-21
Apply reflection catcher and transparent background in viewport in Full mode

ardraeiss:
### PURPOSE
Added support for Reflection catcher and Transparent background in Viewport in Full mode.

### EFFECT OF CHANGE
- added support for Reflection catcher and Transparent background in Viewport in Full mode;

### TECHNICAL STEPS
- created and used background filter in viewport;
- resolved required AOVs when background filter is active;
- used Color AOV for first render iteration.

### NOTES FOR REVIEWERS
- first iteration is ignored since AOVs aren't properly calculated yet so the result is nearly completely black. The color AOV is displayed instead so the user can see where she is during scene editing and camera movement.


bnagirniak (approved):
Tested - works ok.

PR #277, opened 2021-05-26
Always add subdivision if displacement enabled

bsavery:
### PURPOSE
Default adding a displacement looks jaggy unless adding subdivision as well

### EFFECT OF CHANGE
Subdivision is enabled by default if displacement is used.  Subdivision is set to a level which is capped at 1 poly per pixel.  This may be too heavy (slow) so if users enable the RPR subdivision object setting they can control it explicitly.


ardraeiss (approved):
OK

src/rprblender/export/mesh.py
```diff
@@ -242,6 +242,14 @@ def assign_materials(rpr_context: RPRContext, rpr_shape: pyrpr.Shape, obj: bpy.t
         if mat.cycles.displacement_method in {'DISPLACEMENT', 'BOTH'}:
             rpr_displacement = material.sync(rpr_context, mat, 'Displacement', obj=obj)
             rpr_shape.set_displacement_material(rpr_displacement)
+            # if no subdivision set that up to 'high' so displacement looks good
+            # note subdivision is capped to resolution
+            if rpr_shape.subdivision is None:
```
ardraeiss, line 247: The question is: what if object is already subdivided by Blender?

bnagirniak (approved):
Looks ok

Issue #270, opened 2021-05-16
builds v3.1.6 doesn't work with Blender 2.93b on Mac OS


Aliprandini:
Blender 2.93 beta, Mac OS 11.3.1, RPR 3.1.6, RX 5700XT

Couldn't activate RPR in Blender. A lot of errors show up. Smth should be wrong with this build for Blender Mac version. Please, re-check this.

bsavery:
actually @ARozghon.... only on linux does it use system python I think?

bsavery:
Meaning they don't need to update the python unless they are building themselves

ARozghon:
yes, that's true, need to update Python to 3.9 on Linux. and that's good if it not need to do on macOS and Windows

Aliprandini:
3.1.10 DOESN'T WORK again

bnagirniak:
@Aliprandini please provide plugin log.

Aliprandini:
Sorry, the RPR could be activated, but before this everybody should delete "module" folder at ~i/Library/Application Support/Blender/2.93/scripts fore Numpy rebuild. 

But I found another problem. Botocore library is downloading every time while Blender.app start while RPR is activated. 

![123](https://user-images.githubusercontent.com/73755453/119254679-b8032800-bbc8-11eb-88ad-d2fe8a222c49.png)

bnagirniak:
Thank you for log and issue with boto library installation, we'll look it closely and we'll fix it.

bnagirniak:
Should be fixed with #293

PR #278, opened 2021-05-27
RPRBLND-1734: Issue in starting plugin when Boto3 failed to install.

bnagirniak:
### PURPOSE
When Boto3 failed to install plugin starts too long every time, because it tries to install Boto3 everytime.
This is fix for #270.

### EFFECT OF CHANGE
Fixed issue with installing Boto3

### TECHNICAL STEPS
Implemented behavior: if Boto3 failed to install next try to install will be after 5 days.
This is done by saving next date of trying to install boto3 into pip_check.txt file in plugin dir.


bsavery:
So was the main issue that boto3 on python 3.9 wasn't available?

bnagirniak:
> 
> So was the main issue that boto3 on python 3.9 wasn't available?

It is available in on python 3.9, but seems on user's machine there are problems with installing it. Maybe because of some rights to install issues.

Issue #245, opened 2021-04-02
This blend file with RPR Blender addon 3.1.0 - possible mem leak leading to eventual system hang


Jeducious:
Ok, so, after testing 3.1.0 noticed that if the Adapt Viewport Resolution and Viewport Denoising and Upscaling options are set, then the RAM consumed by the parent blender process begins to increase as the viewport is updated and re-rendered, eventually leading to a complete system hang. 

Turning off these options the RAM usage is stable. 

I started Blender a second time and this time turned off the two options mentioned above, then changed my view to the material preview. 

All was fine, then I turned on just the Viewport Denoising and Upscaling option. I got an immediate RAM spike which I had not had using the system without this option in the rendered viewport, with no RAM increases so far, and plenty of rendering in the viewport done. 
<img width="1440" alt="RAM usgae increase when using Viewport Denoising and Upscaling" src="https://user-images.githubusercontent.com/34228929/113459537-92c42a00-9461-11eb-9feb-70624789e83d.png">

[sierpinski Tetrahedra_3.blend.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/files/6251471/sierpinski.Tetrahedra_3.blend.zip)

Jeducious:
Update, switched from material preview to rendered view, pretty much an immediate system hang. 

System works fine so long as I don't enable the Viewport Denoising and Upscaling option

bnagirniak:
Should be fixed with #280

PR #280, opened 2021-05-29
RPRBLND-1719: Possible memory leak with viewport denoising

bnagirniak:
### PURPOSE
Upscale filter in viewport isn't freeing correctly. Bug #245 

### EFFECT OF CHANGE
Fixed memory leak in viewport with upscale_filter.

### TECHNICAL STEPS
Added freeing upscale_filter in stop_render().
Improvement: added resolve() paths in ML and upscale filters.


ardraeiss (approved):
OK

bsavery (changes requested):
Your description says there is some freeing that happens in stop render.  I don't see this part of the change?  Am I missing something?

bnagirniak:
> 
> 
> Your description says there is some freeing that happens in stop render. I don't see this part of the change? Am I missing something?

It happens in line: self.upscale_filter = None

I also found memory leaks for RPR1 and hybrid modes and added fixes in last commit.

PR #284, opened 2021-06-08
Used frame Position for Motion Blur

ardraeiss:
### PURPOSE
Cycles have the Position option in the Motion Blur settings to choose which part of frame use for calculations.
This PR adds usage of this option to the RPR.

### EFFECT OF CHANGE
- added support for Position in Motion Blur settings.

### TECHNICAL STEPS
- added Motion Blur frame Position from Cycles to the UI;
- used frame Position to set subframe for next;
- used the same subframe for objects and instances sync to keep Cycles-like behavior;
- restored original scene subframe 0.0 after objects and instances synchronization;
- added more checks for render cancellation to the particles sync.


src/rprblender/engine/engine.py
```diff
@@ -79,10 +79,17 @@ def depsgraph_instances(self, depsgraph: bpy.types.Depsgraph):
...
 
     def cache_blur_data(self, depsgraph: bpy.types.Depsgraph):
-        cur_frame = depsgraph.scene.frame_current
+        scene = depsgraph.scene
+        cur_frame = scene.frame_current
+
+        subframe = {  # following Cycles behavior
+            'START': 1.0,  # shutter opens at the current frame, means at the end
+            'CENTER': 0.5,
+            'END': 0.0,  # shutter closes at the current frame, means at the start
```
bsavery, line 88: So if the shutter is closing at the current frame shouldn't we blur from 
n-1 -> N
and start would be 
N-> N+1
and center would be N -.5 -> N + .5

ardraeiss: Yes, that looks closer to Cycles. Done.

bnagirniak (approved):
Code looks ok.

bsavery (approved):

PR #288, opened 2021-06-10
iterate through all layers to bake objects.

### LINKED ISSUE #274: Bake all Unsupported Nodes Error / Iteration over deleted objects
Baking all unsupported nodes fails on standard .blend file.

1.  Versions:
Blender 2.92
RPR Version 3.1.6 (newest weekly build at time of writing)

2.   a .blend scene to reproduce this simply. 
the new Blender 2.92 .blend from here
https://cloud.blender.org/p/gallery/60337d495677e942564cce76

## Issue
and
## Steps to Reproduce
1. Download and open the blend file from above
2. In the Render Settings, in the Quality panel there is the Node Baking subsection. Here you will find the Bake All Unsupported Nodes button. Press it.
3.
--> Error Message:
Python: Traceback (most recent call last):
  File "C:\Users\diete\AppData\Roaming\Blender Foundation\Blender\2.92\scripts\addons\rprblender\operators\nodes.py", line 135, in execute
    obj.select_set(True)
RuntimeError: Error: Object 'Cube' can't be selected because it is not in View Layer 'View Layer'!

-- My thoughts:
The "Cube" had been deleted a long time ago. You guys are iterating not only over all active view objects but you are instead iterating over all ever existed objects. Then you want to do something to thoses deleted objects, which obviously must fail.
No good.

## System Setup (Environment)

Win 10

Nvidia 2070 super

2.92
RPR 3.1.6.

bsavery:
### PURPOSE
Currently the "bake all nodes" function does not work with multiple view layers because it selects objects as part of the baking process.  Selecting an object not in the view layer throws a blender error.

### EFFECT OF CHANGE
Iterate over view layers when baking all objects.  Fixes #274


src/rprblender/operators/nodes.py
```diff
@@ -110,37 +110,43 @@ def poll(cls, context):
...
-        selected_object.select_set(True)
+                    old_selection = obj.select_get()
+                    obj.select_set(True)
+                    bake_nodes(nt, nodes_to_bake, material_slot.material, int(resolution), obj)
+                    obj.select_set(old_selection)
 
-        return {'FINISHED'}      
+                    baked_materials.append(material_slot.material.name)
+ 
+        selected_object.select_set(True)
```
ardraeiss, line 147: There could be an issue with selected object from different ViewLayer if object is disabled in the current ViewLayer. It will look like this:
Error: Python: Traceback (most recent call last):
  File "D:\Work\RPR\blender_review\RadeonProRenderBlenderAddon\src\rprblender\operators\nodes.py", line 147, in execute
    selected_object.select_set(True)
RuntimeError: Error: Object 'Suzanne' can't be selected because it is not in View Layer 'View Layer.001'!

ardraeiss (approved):
OK

bnagirniak (approved):
Tested - no issue anymore which is described in #274

PR #294, opened 2021-06-20
Improve prop name readability in Object visibility UI

rsgnz:
rpr_visibility_grid

### PURPOSE
Improve prop name readability in Object visibility UI
Switch from grid to vertical list to avoid ellipsis in the props names.
Set the same prop order as in Cycles.

### EFFECT OF CHANGE

rpr_visibility_column


bnagirniak (approved):
Ok from my side

PR #297, opened 2021-06-24
RPRBLND-1755: Math nodes in materials cause a crash

VascoPi:
### PURPOSE
The scenes below can't be rendered with the message
ValueError: ('Incorrect math operation', 'SMOOTH_MIN')
ValueError: ('Incorrect math operation', 'COMPARE').

### EFFECT OF CHANGE
Added support for Compare, Smooth minimum and Smooth maximum operations for Math Node.

### TECHNICAL STEPS
1. Added support for "COMPARE", "SMOOTH MINIMUM" and "SMOOTH MAXIMUM" math operations.
2. Changed Raise ValueError to log.warn when math operation not supported


bnagirniak (approved):
Tested - works good.

PR #299, opened 2021-06-25
RPRBLND-1742: Add separate settings for Outline Render

VascoPi:
### PURPOSE
Add Samples option for Outline Rendering
Setup for Samples default value of 16
Render Properties -> Sampling -> Outline Samples

### EFFECT OF CHANGE
Added Samples option for Outline Rendering
Samples option of Outline Rendering is separated from render max samples option
By default value it is 16
Render Properties -> Sampling -> Outline Samples

### TECHNICAL STEPS
Added property RPR_RenderLimits.contour_render_samples
Added ui item at RPR_RENDER_PT_limits
Added variable in RenderEngine._render_contour
Changed in RenderEngine._render_contour, render_samples to contour_render_samples


bnagirniak (approved):
Tested - works good.

bsavery:
I am not sure exactly if I have an answer here. But we do normally put the sample settings in the render settings.  These are in the view layer settings.  Would it make more sense to put them in the render settings?

bnagirniak:
We discussed it, I suggested to place contour samples it in Contour settings panel, to be more close to other related settings. If you think better to move it into render settings panel, lets do this. @bsavery what do you prefer?

bsavery:
I think I prefer a section with other samples in render settings.

bnagirniak (approved):

bsavery (approved):

PR #305, opened 2021-06-30
RPRBLND-1790: Error when doing viewport render with cloud scene

VascoPi:
### PURPOSE
Expected the viewport renderer to render the scene
But blender is complaining that the mesh object emitter hasn't been evaluated (error comes from rna_ParticleSystem_uv_on_emitter).
Possible solutions would be to evaluate hair instances last or do some try catch around this.

### EFFECT OF CHANGE
Viewport renderer handles the scene

### TECHNICAL STEPS
Changed check for modifier visability, modifier.show_viewport and modifier.show_render


bnagirniak (approved):
Looks ok

PR #314, opened 2021-07-06
RPRBLND-1741: Add Outline render output for Render Layer Node in Compositor

VascoPi:
### PURPOSE
Need to add Outline render output for Compositor that will be useful for users and will add more options for render results.

### EFFECT OF CHANGE
Outline pass appears in RenderLayers node when Full quality render option and Outline Rendering are enabled.

### TECHNICAL STEPS
Added Outline pass registration update.
Outline pass appears is in RenderLayers node when Full quality render option and Outline Rendering are enabled.


bnagirniak (approved):
Tested by QA - works good.

PR #324, opened 2021-07-13
RPRBLND-1788: Image sequence in Image texture node not working correct.

VascoPi:
### PURPOSE
Load image sequence in ShaderNodeTexImage and get correct result in Render and Viewport.
Investigate work of Cyclic and Auto Refresh options, and make them work as expected.

### EFFECT OF CHANGE
Now Viewport and Render works as expected, compare to blender Image texture node in SEQUENCE mode.
Added support for Cyclic and Auto Refresh options in Image texture node.
Added support for also only numeric Image filenames for Image texture node .

### TECHNICAL STEPS
Despgraph doesn't provide updates in Viewport for ShaderNodeTexImage with activated Auto Refresh option.
Added force update for Viewport render for materials with ShaderNodeTexImage with SEQUENCE source.
Added calculations for frame_current if use_auto_refresh is off, due to blender provides incorrect value in some cases.
Fixed get_sequence_frame_file_path to get correct filepath also if filename contains only digits.
Removed constant for max zeros in filename, now it takes from the filename.


src/rprblender/engine/viewport_engine.py
```diff
@@ -554,7 +554,22 @@ def sync_update(self, context, depsgraph):
         updates = []
         for obj_type in (bpy.types.Scene, bpy.types.World, bpy.types.Material, bpy.types.Object,
                          bpy.types.Collection, bpy.types.Light):
-            updates.extend(update for update in depsgraph.updates if isinstance(update.id, obj_type))
+            for update in depsgraph.updates:
+                if isinstance(update.id, obj_type):
+                    updates.append((update.id, update.is_updated_geometry, update.is_updated_transform))
+
+        # despgraph doesn't provide updates for ShaderNodeTexImage with activated Auto Refresh option
```
bnagirniak, line 561: Getting materials with ShaderNodeTexImage node and source == 'SEQUENCE' has to be optimized. It should be used only when we have Scene object in updates and frame was changed.

bnagirniak (changes requested):
Optimization is required.

src/rprblender/engine/viewport_engine.py
VascoPi: Done in new commit.

bnagirniak (approved):
Tested - works ok

PR #325, opened 2021-07-13
RPRBLND-1800: "KeyError 10" with Shadow Catcher activation in viewport render

VascoPi:
### PURPOSE
Activate shadow catcher without error.
Fix an issue get KeyError 10 with shadow catcher activation. when Viewport render starts with activated reflection catcher.

### EFFECT OF CHANGE
Fixed an issue get KeyError 10 with shadow catcher activation.

### TECHNICAL STEPS
Removed sync_catchers(). It makes no effect at sync_update, due to RPRContex2 doesn't fully provide this method.

### NOTES FOR REVIEWERS
There's still an issue with shadow catcher and reflection catcher activation.
Even before fix, switching the options in Viewport in realtime, the result isn't always correct.
Kind of some framebuffer issue or shadow catcher and reflection catcher sync.
Need more investigation.


bnagirniak (approved):
Tested - no error anymore.

PR #328, opened 2021-07-15
RPRBLND-1739: Improve "hide in viewport" option

VascoPi:
### PURPOSE
Improve the "hide in viewport" option, because it not hide objects correctly in some cases

### EFFECT OF CHANGE
Added viewport update for hide/unhide object
Added viewport update for create/delete object

### TECHNICAL STEPS
Changed selection check in sync_update, because of, some updates, like object create/remove or object hide, are ignored
Added viewport update for hide/unhide object
Added viewport update for create/delete object


bnagirniak (approved):
Looks good.

src/rprblender/engine/viewport_engine.py
```diff
@@ -556,6 +551,11 @@ def sync_update(self, context, depsgraph):
                          bpy.types.Collection, bpy.types.Light):
             updates.extend(update for update in depsgraph.updates if isinstance(update.id, obj_type))
 
+        if context.selected_objects != self.selected_objects and not updates:
+        # only a selection change
```
bsavery, line 555: really minor thing but comment should be indented.

bsavery (changes requested):
Sorry, minor nit pick.

PR #335, opened 2021-07-19
RPRBLND-1644: Principled BSDF Emission Strength not working with RPR

VascoPi:
### PURPOSE
Make Emission Strength in Principled BSDF work as expected. Add support for a value >1.

### EFFECT OF CHANGE
Added Emission Strength support for Principled BSDF.

### TECHNICAL STEPS
Added Emission Strength param support for ShaderNodeBsdfPrincipled, now RPR supports a value >1.
Now MATERIAL_INPUT_UBER_EMISSION_COLOR takes into account emission and emission strength.


bsavery (commented):
So this is more of a question than a review.  I'd like to see you justify the math here.  Internally RPR does emission color * emission strength.  

So as far as I can see, we will actually multiply the emission color * emission strength twice here?  

Secondly emission seems to be minimum of 1?  what if strength is .5?

VascoPi:
> So this is more of a question than a review. I'd like to see you justify the math here. Internally RPR does emission color * emission strength.

As I see ShaderNodeBsdfPrincipled uses these two inputs for emission part of the shader.
MATERIAL_INPUT_UBER_EMISSION_WEIGHT - emission_weight, transition between emission and diffuse (0.0 to 1.0), for example at a value 0.0 - 0.5 material still catches some shadows/reflections and so on. It's clamped internally, so the max value is 1.0. Same in Uber Shader.
So I decoded to use color input also for emission_sterngth.
MATERIAL_INPUT_UBER_EMISSION_COLOR - emission * emission_strength. I make the math to input color value and emission strength.

>So as far as I can see, we will actually multiply the emission color * emission strength twice here?

Yes, thanks, will fix it.

src/rprblender/nodes/blender_nodes.py
```diff
@@ -818,6 +818,7 @@ def enabled(val: [NodeItem, None]):
             transmission_roughness = self.get_input_value('Transmission Roughness')
 
         emission = self.get_input_value('Emission')
+        emission_strength = self.get_input_value('Emission Strength')
```
bnagirniak, line 821: Add `if enabled(emission):` and blender version check here

src/rprblender/nodes/blender_nodes.py
```diff
@@ -896,8 +897,9 @@ def enabled(val: [NodeItem, None]):
         # Emission -> Emission
         if enabled(emission):
             # more related formula for emission weight:
-            emission_weight = emission.average_xyz().min(1.0) * 0.5 + 0.5
-
+            emission *= emission_strength
+            emission_weight = (emission).average_xyz().min(1.0) * 0.5 + 0.5
```
bnagirniak, line 901: just emission.average_xyz

bnagirniak (changes requested):
Optimization and blender version check is required

bnagirniak (approved):
Tested - works good

bsavery:
@ARozghon please test functionality

DagerD (approved):
Tested. Works ok.

ARozghon:
approve from QA, all works fine and we get the correct autotest results

PR #336, opened 2021-07-21
RPRBLND-1832: Motion blur isn't applicable to camera motion with position End of Frame

bnagirniak:
### PURPOSE
Motion blur isn't applicable to camera motion with position End of Frame + there are other issues with setting required frame.

### EFFECT OF CHANGE
Fixed issues with motion blur for frame positions 'Center of Frame', 'End of Frame'.

### TECHNICAL STEPS
1. Fixed cache_blur_data() with correct setting scene frame.
2. Fixed motion blur set in scene export in RenderEngine and ExportEngine.


src/rprblender/engine/engine.py
```diff
@@ -83,28 +83,25 @@ def cache_blur_data(self, depsgraph: bpy.types.Depsgraph):
         position = scene.cycles.motion_blur_position
 
         if position == 'END':  # shutter closes at the current frame, so [N-1 .. N]
-            start_frame = scene.frame_current - 1
-            subframe = 0.0
+            from_frame = (scene.frame_current, 0.0)
```
bsavery, line 86: wouldn't this make motion backwards?

bnagirniak: Yes it would, but seems this is exactly what user expects for this option

VascoPi:
Tested, works fine.
Render as expected with all three options of motion blur "START", "MIDDLE", "END".

VascoPi (approved):

PR #337, opened 2021-07-23
Updated settings in min_samples and max_samples render properties to …

bnagirniak:
### PURPOSE
There is request from other devs/artists to have possibility to render scene with less then 16 iterations.

### EFFECT OF CHANGE
Updated limits for min and max samples render properties to: min=1, soft_min=16.

### TECHNICAL STEPS
Updated limits for min_samples and max_samples render properties to: min=1, soft_min=16.


VascoPi (approved):
Works fine.

PR #338, opened 2021-07-23
RPRBLND-1645: Create operator to export denoiser training data

VascoPi:
### PURPOSE
Add an operator (no ui) to export all training data for a scene. 
Export for each camera in scene.
Export for sample in 1, 2, 4, 8, 4096.
Render color, normal, depth, albedo aovs to disk with format {cam_name}.{aov_name}.{samples}.bin. Save as "exr" file format 32 bit

### EFFECT OF CHANGE
Added a new operator at bpy.ops.rpr.export_training_data, which allows to export render data form each camera to files.

### TECHNICAL STEPS
Added export_training_data.py at operators module
Added and registered  new operator class RPR_EXPORT_OP_export_training_data,
Create rpr_context with ExportEngine, ExportEngine2
Added predefined export of:
      AOV_COLOR
      AOV_DIFFUSE_ALBEDO
      AOV_DEPTH
      AOV_SHADING_NORMAL
      AOV_VARIANCE
      AOV_DIRECT_DIFFUSE
      .AOV_DIRECT_ILLUMINATION
      AOV_DIRECT_REFLECT
      AOV_INDIRECT_DIFFUSE
      AOV_INDIRECT_ILLUMINATION
      AOV_INDIRECT_REFLECT
      AOV_UV
      AOV_VELOCITY
      AOV_WORLD_COORDINATE
Added export of a frame_buffer to files according to extension exr, bin or other
Added export for frame_range
Added export for every camera in scene

### NOTES FOR REVIEWERS
not included random seen due to CONTEXT_RANDOM_SEED is not supported yet, returns ERROR_UNSUPPORTED = -23
not included fps and motion blur position, leave it to be defined by user


src/rprblender/operators/export_training_data.py
```diff
@@ -0,0 +1,174 @@
...
+        default=1920,
+        name="Width",
+        description="Set render width resolution",
+    )
+    height: IntProperty(
+        default=1080,
+        name="Height",
+        description="Set render height resolution",
+    )
+    samples: StringProperty(
```
bsavery, line 67: bpy.props.IntVectorProperty technically I'd rather have this be an int vector property.  In all honesty though I think this can be hard coded as this is the only set the denoiser needs.

VascoPi: Ok, got it, will change the set.

bnagirniak (approved):
Tested - works good

src/rprblender/operators/export_training_data.py
bnagirniak: We discussed this option. IntVectorProperty is size fixed array. I suggested to make it as StringProperty, becuase it gives us more flexibility in setting samples including testing.
@bsavery if you think better to use IntVectorProperty or hardcode this, let it be.

PR #345, opened 2021-08-19
More generic call of python3

Toorero:
### PURPOSE
For people that have python3.9 installed (e.g. Archlinux) and all the build process breaks because BlenderHelper doesn't get build for some reason as well as `python3.7` command doesn't even exist.

### EFFECT OF CHANGE
Build starts on all python3 versions, but may break if not using python3.7.

### NOTES FOR REVIEWERS
I think it's better to not arbitrary enforce a python version. I rather like to try to compile it with python 3.9 and see it break then to change all the files, so I can even compile it on python3.9. If it breaks, I know I should probably install python3.7


Toorero:
Also added another `cffi_libs_dir` location where `cffi` is stored on my system.

bsavery:
I'm not opposed to this.  Ideally maybe the better thing would be to build directly from the version of python that blender is using, however the built in python in blender doesn't have all the build tools usually.  

Also I would note there is build_osx.sh

Toorero:
I don't quite get what you mean.

Python is a dependency of Blender that does not imply it's bundled with Blender and I would assume a lot of distros would entangle this bundle even if it would exist in the first place (like with Native-Steam) because it causes a lot of troubles, adds redundancies and is a security concern. [Blender is also based on python3.9](https://developer.blender.org/diffusion/B/browse/master/build_files/build_environment/cmake/versions.cmake$192), so it's even more staggering for me why one would force using python3.7 especially if building with python3.9 works flawless. 

I don't know anything about OSX, so it would be cool if you could tell me more about the OSX and Linux connection. As far as I know, there isn't even an official package manager system on OSX.

bsavery:
Keep in mind I'm talking about the "bundled python" on MacOS and windows.  The reason we build with python 3.7 by default is because < Blender 2.93 it uses 3.7.  And we need to support both versions.

Toorero:
Yeah, then adjust the build files for Windows and macOS to use the build in Python version but I don't see the connection of the build process of Linux and macOS/Windows.

Toorero:
Oh so it's not the build FOR Linux but rather a build on Linux? So the plugin itself is platform independent? Nonetheless, this build will fail like this on many systems running modern Linux.

Toorero:
Ah and if you want to figure out the python version Blender is "using" you can just run
```
curl -s https://raw.githubusercontent.com/blender/blender/blender-v{VERSION}-release/build_files/build_environment/cmake/versions.cmake | sed -nr "s/set\(PYTHON_SHORT_VERSION (.*)\)/\1/p"
```

Toorero:
Also, if you say you want to support Blender 2.80 (with python3.7) but in the meantime Blender 2.93 uses python3.9 you have to code your python code in the way that it runs on python3.7 and python3.9 anyway, so there shouldn't be a problem to just use the latest version, should it? Otherwise, there have to be Blender-specific version builds, which doesn't seem like there are any according to the release page.

Also keep in mind that if you want to build Blender for yourself the distro manager also has to provide the right python3 version anyway, so they are basically doing the job of having the right python version ready for Blender and therefore for the build process as well (if Blender is installed).

bnagirniak (changes requested):
We can't approve proposed changes because:
1. We support Blender 2.92 and earlier versions, which uses python 3.7 and Blender 2.93+ with python 3.9. Therefore build.py has to be called with both pythons. Current changes breaks our build and test process.
2. We have 'develop' branch and PRs have to be created with changes into that branch. 'master' branch is only for releases.

We'll take into consideration your request of plugin build. Overall our build process has to be improved, we can do it in future.

Toorero:
Okay now I also get why RPRBlenderHelper is only build for python 3.7, so it won't be build twice. So if you want to build for Blender 3.93+ you still have to install python3.7...

Toorero:
Then I would propose to add a buildflag to buid.py to skip building RPRBlenderHelper, so that `python3 build.py` actually works and in the `build.sh` and the other respective files use the flag to skip the building in the 3.9 call of `build.py`. Does this sound acceptable?

orsonteodoro:
There are more python3.7 references in the BlenderPkg folder.

bnagirniak:
Closing this PR as no more updates were made.

PR #351, opened 2021-09-06
RPRBLND-1759: Blender RPR 2 Hetero Volumes

VascoPi:
### PURPOSE
Northstar has a new API for using heterogenous volumes. Integrate to allow smoke and fog rendering from Blender .

### EFFECT OF CHANGE
Heterovolume now supports in Full quality (Northstar)

### TECHNICAL STEPS
Heterovolumes in Northstar works mainly using shading.
Now it creates a volume mesh from a smoke domain object at mesh.sync and apply volume material.
Added support of heterovolume for ShaderNodeVolumePrincipled and ShaderNodeVolumeScatter nodes with export_rpr2.
Added support for Grid object input MaterialNodeSetInputGridDataByKey.
Added heterovolume simulation update viewport_engine.py , so additionally volume shaderupdates after volume mesh due to 
volume shader uses density grid.

### NOTES FOR REVIEWERS
When press playback, volume simulation updates only at stop, if manually change frame or volume shader input is animated everything is ok.
Lookup tables and Volume node info maybe move to another task so it needs more work to do
Not tested support of VDB files, maybe move to another task.


src/rprblender/export/mesh.py
```diff
@@ -322,8 +324,20 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, **kwargs):
...
 
+    transform = object.get_transform(obj)
     deformation_data = rpr_context.deformation_cache.get(obj_key)
-    if deformation_data and np.any(data.vertices != deformation_data.vertices) and \
+
+    if smoke_modifier and isinstance(rpr_context, RPRContext2):
+        transform = volume.get_transform(obj)
+        rpr_shape = rpr_context.create_mesh(
+            obj_key,
+            data.vertices, data.normals, data.uvs,
```
bnagirniak, line 334: Do we need to provide mesh data here with MESH_VOLUME_FLAG?

bnagirniak (commented):
Tested - works good, one question about providing mesh data for volume object.

src/rprblender/export/mesh.py
VascoPi: Yes, it can be optimized. Fixed with creating an empty Mesh object.

bnagirniak (approved):
Tested - works good.

DagerD (approved):
Tested. Works ok.

PR #356, opened 2021-09-13
RPRBLND-1905: Export to RPR provide crash

VascoPi:
### PURPOSE
Export to rpr provide blender crash.

### EFFECT OF CHANGE
Export scene as expected, without crash.

### TECHNICAL STEPS
Fixed compatibility passing NULL value for rprsCtx argument.


bnagirniak (approved):
Tested - works good

DagerD (approved):
Tested. Works ok.

PR #360, opened 2021-09-16
RPRBLND-1670: Hair inside particles not working

VascoPi:
### PURPOSE
Add an object with hair to particles.

### EFFECT OF CHANGE
Render hair on particles.

### TECHNICAL STEPS
Added hair sync to instance object.
Added check for instance object in hair module.
Modified hair_key for viewport render update.
Added new transform matrix if the hair is an instance.

### NOTES FOR REVIEWERS
Materials update in Viewport render is not always correct. The final render works well.


bnagirniak (approved):
Looks good, made some code improvements

DagerD (approved):
Tested. Works ok.

PR #370, opened 2021-10-04
RPRBLND-1896: Case scene.world == None isn't handled by plugin.

DagerD:
### PURPOSE
There is errors in console during rendering and UI operations if we delete world light from scene.

### EFFECT OF CHANGE
No more errors if world light doesn't exist.

### TECHNICAL STEPS
Added checks for world light everywhere it is needed

### DIFF
--- a/src/rprblender/engine/render_engine.py
+++ b/src/rprblender/engine/render_engine.py
@@ -672,7 +672,7 @@ class RenderEngine(Engine):
             self.camera_data.export(rpr_camera)
 
         # Environment is synced once per frame
-        if scene.world is not None:
+        if scene.world:
             if scene.world.is_evaluated:  # for some reason World data can came in unevaluated
                 world_data = scene.world
             else:
--- a/src/rprblender/engine/render_engine.py
+++ b/src/rprblender/engine/render_engine.py
@@ -672,12 +672,13 @@ class RenderEngine(Engine):
             self.camera_data.export(rpr_camera)
 
         # Environment is synced once per frame
-        if scene.world.is_evaluated:  # for some reason World data can came in unevaluated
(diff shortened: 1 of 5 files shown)


src/rprblender/ui/world.py
```diff
@@ -24,13 +24,19 @@ class RPR_WORLD_PT_environment(RPR_Panel):
     bl_context = 'world'
 
     def draw_header(self, context):
```
bnagirniak, line 26: override poll()

src/rprblender/ui/world.py
```diff
@@ -163,9 +182,15 @@ class RPR_WORLD_PT_sun_sky(RPR_Panel):
 
     @classmethod
     def poll(cls, context):
+        if context.scene.world is None:
+            return super().poll(context)
```
bnagirniak, line 186: return False

src/rprblender/engine/render_engine.py
```diff
@@ -672,12 +672,13 @@ def sync(self, depsgraph):
             self.camera_data.export(rpr_camera)
 
         # Environment is synced once per frame
-        if scene.world.is_evaluated:  # for some reason World data can came in unevaluated
-            world_data = scene.world
-        else:
-            world_data = scene.world.evaluated_get(depsgraph)
-        world_settings = world.sync(self.rpr_context, world_data)
-        self.world_backplate = world_settings.backplate
+        if scene.world is not None:
```
bnagirniak, line 675: simplify to: `if scene.world:`

bnagirniak (changes requested):

DagerD:
Close this PR due to wrong target branch. Will create new.

PR #371, opened 2021-10-04
RPRBLND-1896: Case scene.world == None isn't handled by plugin.

DagerD:
PURPOSE
There is errors in console during rendering and UI operations if we delete world light from scene.

EFFECT OF CHANGE
No more errors if world light doesn't exist.

TECHNICAL STEPS
Added checks for world light everywhere it is needed

### DIFF
--- a/src/rprblender/engine/preview_engine.py
+++ b/src/rprblender/engine/preview_engine.py
@@ -110,10 +110,10 @@ class PreviewEngine(Engine):
         camera.sync(self.rpr_context, depsgraph.objects[depsgraph.scene.camera.name])
 
         # export world only if active_material.use_preview_world is enabled
-        preview_obj = next(obj for obj in self.depsgraph_objects(depsgraph)
-                               if obj.name.startswith('preview_'))
-        if preview_obj.active_material and preview_obj.active_material.use_preview_world:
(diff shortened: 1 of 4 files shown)


VascoPi (changes requested):
Still got the issue in Legacy, High, Mid, Low modes, while Viewport render.

bnagirniak (changes requested):
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "D:\amd-gpuopen\RadeonProRenderBlenderAddon\cmd_tools/test_rpr.py", line 32, in <module>
    import rprblender
  File "D:\amd-gpuopen\RadeonProRenderBlenderAddon\src\rprblender\__init__.py", line 50, in <module>
    from .engine.preview_engine import PreviewEngine
  File "D:\amd-gpuopen\RadeonProRenderBlenderAddon\src\rprblender\engine\preview_engine.py", line 118
    self.rpr_context.enable_aov(pyrpr.AOV_COLOR)
    ^
IndentationError: expected an indented block
```

bnagirniak (approved):
Tested - works good

VascoPi (approved):
Looks good, and World Properties panel behavior is correct.

PR #372, opened 2021-10-05
RPRBLND-1929: Material override is ignored on export

VascoPi:
### PURPOSE
Material override is ignored on export.

### EFFECT OF CHANGE
Material override now takes into account when export rpr.

### TECHNICAL STEPS
Passed material_override into object and instances sync methods.


bnagirniak (approved):
Looks good

DagerD (approved):
Tested. Works ok.

PR #374, opened 2021-10-05
RPRBLND-1908: Add an option for the RPR_AOV_CAMERA_NORMAL in RPR Passes in GUI.

DagerD:
### PURPOSE
Need to add option RPR_AOV_CAMERA_NORMAL  in RPR Passes in GUI.

### EFFECT OF CHANGE
RPR passes in View Layer now contain Camera Normal option

### TECHNICAL STEPS
aovs_info extended with new property


bnagirniak (approved):
Tested - works good.

VascoPi (approved):
Tested, works well. Provides correct results in Compositor and Image Editor.

PR #375, opened 2021-10-05
RPRBLND-1925: Point light difference between cycles and rpr.

bnagirniak:
### PURPOSE
There is a huge difference by point light behavior between cycles and rpr.

### EFFECT OF CHANGE
Fixed light power difference between cycles and RPR.

### TECHNICAL STEPS
Found suitable coefficients for power adjustments: 0.01 for point light power and 0.1 for are light.

### NOTES FOR REVIEWERS
Experimented with suitable coefficients for default units in light power to make RPR light power be almost the same as in cycles.


DagerD (approved):
Tested. Looks much better.

src/rprblender/export/light.py
```diff
@@ -45,7 +44,7 @@ def get_radiant_power(light: bpy.types.Light, area=0.0):
     if light.type in ('POINT', 'SPOT'):
         units = rpr.intensity_units_point
         if units == 'DEFAULT':
-            return color * math.sqrt(light.energy)  # to match cycles
+            return default_intensity * 0.01     # to match cycles
```
bsavery, line 47: Do we have any justification for this number?

bsavery (commented):
Just curious where this number came from.

src/rprblender/export/light.py
bnagirniak: Experimented with different coefficients on different scenes. Tried to find more suitable.

bsavery:
> Just curious where this number came from.

Please note in comment this was a number determined with experimentation.  Also to confirm this works well for different "size" point lights and different scenes @ARozghon

bsavery:
Waiting for review from @ARozghon to merge.

PR #376, opened 2021-10-05
Rprblnd 1926

DagerD:
### PURPOSE
Mute node doesn't work if input and output sockets have different types

### EFFECT OF CHANGE
Muting works properly in such cases

### TECHNICAL STEPS
Changed comparison between wrong sockets


VascoPi (changes requested):
Works well if a muted node has the same type (float, vector, ...) of output as the previous node, compare to Cycles.

bnagirniak (approved):
Tested - works good

PR #381, opened 2021-10-18
RPRBLND-1922: Liquid simulation not rendered with RPR

DagerD:
### PURPOSE
Liquid simulation stopped working after hetero volumes PR #351 

### EFFECT OF CHANGE
Objects with Fluid modifier now renders correctly.

### TECHNICAL STEPS
Added check for material existence on object


VascoPi (changes requested):
Heterovolume is broken now.
If possible maybe it's better to avoid a check with the material sockets.

VascoPi:
Please look at volume.py to get_smoke_modifier:
Try this check.
return next((modifier for modifier in obj.modifiers
                     if modifier.type == 'FLUID' and modifier.fluid_type == 'DOMAIN'
                     and modifier.domain_settings.domain_type == 'GAS'), None)

PR #389, opened 2021-10-27
Toon color input should always be shown

bsavery:
### EFFECT OF CHANGE
Toon shader always shows the input color and sets on the node.


bnagirniak (approved):
Looks good

ARozghon (approved):
tested and all works as expected

PR #406, opened 2021-11-25
RPRBLND-2011: Blender 3.0 support.

bnagirniak:
### PURPOSE
Blender 3.0 is available.

### EFFECT OF CHANGE
Made plugin to work with Blender 3.0+

### TECHNICAL STEPS
Adjusted visibility settings for Blender 3.0

### NOTES FOR REVIEWERS
Installed plugin works good. 
BUT running plugin in developer mode produces errors in console log `Windows fatal exception: access violation`. This is due to some Windows access issues when calling Core dll functions. To fix this relates to separate task.


VascoPi (approved):
Works well, tested.

PR #416, opened 2021-12-07
Update fetching Python executable path

PahiPohaku:
Because bpy.app.binary_path_python was removed during Blender 2.9 and doesn't work anymore, addon activation in preferences fails with AttributeError.

### PURPOSE
Making rpr usable in current and future Blender versions

### EFFECT OF CHANGE
Activating the addon is possible again.

### TECHNICAL STEPS
Replaced reference to bpy.app.binary_path_python with reference to sys.executable

### NOTES FOR REVIEWERS
Nothing further.


bsavery (approved):

bnagirniak (approved):
Ok, thank you for the fix

PR #469, opened 2022-03-22
RPRBLND-2122: Nodes are not removed after clicking "Convert Principled to Uber"

VascoPi:
### PURPOSE
Nodes are not removed after clicking "Convert Principled to Uber".

### EFFECT OF CHANGE
Principled DSDF node is removed after "Convert Principled to Uber".

### TECHNICAL STEPS
added remove function for Principled DSDF node.


DagerD (approved):
Tested. Works ok.

bsavery:
Question.  Why SHOULD the principled node be removed?  I think I wrote the original code and left it in on purpose.

bnagirniak (approved):
Looks good

bsavery:
My original intention was to leave the Principled node in case the user wanted to switch back.

VascoPi:
Not valid.

PR #471, opened 2022-03-24
Map Range node

bsavery:
### PURPOSE
Handle the map range node in linear mode only for now

TODO for users to expand upon

### EFFECT OF CHANGE
No purple materials when using map range.


src/rprblender/nodes/blender_nodes.py
```diff
@@ -1760,6 +1760,32 @@ def export_hybrid(self):
...
+        from_max = self.get_input_value('From Max')
+        to_min = self.get_input_value('To Min')
+        to_max = self.get_input_value('To Max')
+        
+        ''' Doing math like this is actually compiled into a 
+            shader that is executed at runtime. '''
+        from_range = from_max - from_min  
+        to_range = to_max - to_min
+        if self.node.clamp:  # you can access node values like this
+            value = value.clamp(from_min, from_max)
```
VascoPi, line 1782: Please check this:
local variable 'value' referenced before assignment

bnagirniak (approved):
Tested - works ok.

ARozghon (approved):
tested manually and we can to merge in Dev branch

PR #473, opened 2022-03-25
RPRBLND-2027: "Object selection" always fire rendering from 0 spp, there is no property changes

VascoPi:
### PURPOSE
 "Object selection" always fires rendering from 0 spp, there is no property changes.

### EFFECT OF CHANGE
Fixed Object selection starts viewport rendering update.

### TECHNICAL STEPS
Improved condition for object selection checking bpy.types.Scene update.
Fixed unnecessary updates of denoiser and upscaler.


src/rprblender/engine/engine.py
```diff
@@ -128,10 +128,13 @@ def setup_image_filter(self, settings):
...
                 self._enable_image_filter(settings)
 
+            return True
+
         elif self.image_filter:
             self._disable_image_filter()
+            return True
 
-        return True
+        return False
```
bnagirniak, line 137: Logic is not changed in this function, please revert.

src/rprblender/engine/engine.py
```diff
@@ -392,10 +395,13 @@ def setup_upscale_filter(self, settings):
...
                 self._enable_upscale_filter(settings)
 
+            return True
+
         elif self.upscale_filter:
             self._disable_upscale_filter()
+            return True
 
-        return True
+        return False
```
bnagirniak, line 404: Same here, logic is not changed in this function, please revert.

src/rprblender/engine/viewport_engine.py
```diff
@@ -586,8 +586,10 @@ def sync_update(self, context, depsgraph):
             for mat in volume_domain_mat:
                 updates.append((mat, None, None))
 
-        if context.selected_objects != self.selected_objects and not updates:
-            # only a selection change
+        # only a selection change
+        if context.selected_objects != self.selected_objects \
+                and len(updates) == 1 and isinstance(updates[0][0], bpy.types.Scene)\
+                and updates[0][1] == updates[0][2] == False:
```
bnagirniak, line 592: Do you mean here: `updates[0][1] == False and updates[0][2] == False` ?

bnagirniak (changes requested):
Commented in code

src/rprblender/engine/engine.py
VascoPi: It's dedicated to the Outliner window which causes updates when clicking on an object two times.

src/rprblender/engine/viewport_engine.py
VascoPi: Sure, thanks.

DagerD (approved):
Tested. Works ok.

bnagirniak (approved):
Looks good

PR #481, opened 2022-04-07
RPRBLND-2079: Blender crash with AOV and Hybrid Pro

VascoPi:
### PURPOSE
Blender crash with AOV and Hybrid Pro.

### EFFECT OF CHANGE
RPR Passe panel shows only AOVs that are supported by the current Render Quality.

### TECHNICAL STEPS
Added 'hybrindpro' key to aovs_info, describes the support of pass by HybridPro engine.
Improved RPR_VIEWLAYER_PT_aovs, do_register_pass ,  export_aovs with the condition.
Small code improvements.


bsavery:
IMO this should be handled at the RPR / Hybrid pro level, NOT worked around in the plugin code, as this leads to ugly solutions that never get fixed.  

Or at worst done in the RPR python binding for hybrid.

src/rprblender/properties/view_layer.py
```diff
@@ -247,162 +248,194 @@ class RPR_ViewLayerProperites(RPR_Properties):
         {
             'rpr': pyrpr.AOV_COLOR,
             'name': "Combined",
-            'channel': 'RGBA'
+            'channel': 'RGBA',
+            'hybridpro': True
```
bnagirniak, line 252: instead of adding 'hybridpro' option better all list variable with unsupported AOVs in pyhybridpro.py

bnagirniak (changes requested):
Add warning with unsupported AOV in hybridpro

VascoPi:
Oh I understand, I'll make another solution. Reported trace to HybridPro team, I wait for feedback.

VascoPi:
Moved to https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/pull/486

PR #487, opened 2022-04-19
RPRBLND-2040: New Blender 3.0 geometry nodes are working incorrectly

VascoPi:
### PURPOSE
New Blender 3.0 geometry nodes are working incorrectly.
Error with apply override material in blender 3.0 and higher.

### EFFECT OF CHANGE
Improved Geometry Nodes support for Viewport and Final render.
Fixed error with apply override material in blender 3.0 and higher.

### TECHNICAL STEPS
Improved Preview Engine camera object search, in blender 3.0 camera has various names while preview render.
Improved object.key method so it returns the unique result since Geometry Nodes can instantiate objects with the same name and different mesh.
Improved Viewport sync_update method with Geometry Nodes condition for Blender 3.0

### NOTES FOR REVIEWERS
There are issues with instances created via Geometry Nodes.
It is mostly related to Viewport render sync_update.
Final render should work fine now.


src/rprblender/engine/viewport_engine.py
```diff
@@ -1023,8 +1039,10 @@ def update_material_on_scene_objects(self, mat, depsgraph):
             objects = self.depsgraph_objects(depsgraph)
             active_mat = material_override
         else:
+            # Geometry Nodes allawed to apply material via node tree, in that case slot name always ''
```
DagerD, line 1042: allowed

DagerD (approved):
Tested. Works ok.

bnagirniak (approved):
Looks good

ARozghon (approved):
working correctly with Blender 3.0.1 and Blender 3.1.2

PR #501, opened 2022-05-11
RPRBLND-2134: Object Index and Material Index from Object Info are not working

VascoPi:
### PURPOSE
Object Index and Material Index from Object Info are not working.

### EFFECT OF CHANGE
Fixed Material Index and Object Index passes.

### TECHNICAL STEPS
Changed type of input value from int to float.

### NOTES FOR REVIEWERS
Noticed that Material Index for Interactive mode returns incorrect results .
HybridPro team reported that the fix is ready to be merged.


ARozghon (approved):
Now working. The scene for autotests - updated

DagerD (approved):
Tested. Works ok.

PR #502, opened 2022-05-11
RPRBLND-2138: Particles Halo Error

VascoPi:
### PURPOSE
What this PR is made for.

### EFFECT OF CHANGE
Fixed an error occurs if particle render type is Halo.

### TECHNICAL STEPS
Fixed API incompatibility changing passing argument and value for blender 3.0 and upper:
diameter -> radius
1.0 -> 0.5


src/rprblender/export/mesh.py
```diff
@@ -136,7 +136,11 @@ def init_from_shape_type(shape_type, size, size_y, segments):
                 bmesh.ops.create_circle(bm, cap_ends=True, cap_tris=True, segments=segments, radius=0.5)
 
             elif shape_type == 'SPHERE':
-                bmesh.ops.create_uvsphere(bm, u_segments=segments, v_segments=segments, diameter=1.0)
+                if BLENDER_VERSION < '3.0':
```
bsavery, line 139: I don't feel strongly one way or another but you could also do:
`bmesh.ops.create_uvsphere(bm, *{"u_segments":segments, "v_segments":segments, "diameter":1.0 ifBLENDER_VERSION < '3.0' else 'radius':0.5} )

PR #505, opened 2022-05-13
RPRBLND-2143: Incorrect work of Object Info node for Random output

VascoPi:
### PURPOSE
Incorrect work of Object Info node for Random output.

### EFFECT OF CHANGE
Fixed render differences while using Random Output of Object Info node.

### TECHNICAL STEPS
Changed object.key method output so it returns unique object name without randomness using data name as a prefix instead.
Small code improvements.


DagerD (approved):
Tested. Works ok.

ARozghon (approved):
an issue with Info node for Random output and Voronoi random seed fixed with this PR

bnagirniak (approved):
Looks good

PR #506, opened 2022-05-13
RPRBLND-2140: add RPR_CONTEXT_RANDOM_SEED we can set the random seed of the renderer

VascoPi:
### PURPOSE
add RPR_CONTEXT_RANDOM_SEED we can set the random seed of the renderer.

### EFFECT OF CHANGE
Added Seed and feature that allows managing seed value of render noise pattern (same as Cycles).

### TECHNICAL STEPS
Added seed int property.
Added anim_seed bool property.
Added set_random_seed that setting RPR_CONTEXT_RANDOM_SEED parameter.
Added UI element 'Seed' and 'Anim Seed'.

### NOTES FOR REVIEWERS
Implemented for Final render.
Discussed with HybridPro team some differences in behavior Final/Interactive and RPR_CONTEXT_RANDOM_SEED parameter. Probably the feature will be implemented in the future.


DagerD (approved):
Tested. Works ok.

bsavery:
My only comment is this should be an "advanced" option

ARozghon (approved):
seeds and random seeds working fine

VascoPi:
> My only comment is this should be an "advanced" option

Moved 'Seed' to 'Advanced' panel.
image

bnagirniak (approved):
Looks good.

PR #514, opened 2022-05-24
RPRBLND-2137: add Temperature Attribute support

VascoPi:
### PURPOSE
add Temperature Attribute support.

### EFFECT OF CHANGE
Fixed inverted result of Temperature Attribute.
Added support for 'velocity' and  'heat' attributes for Principled Volume Shader.
Fixed error caused by 3 or 4 length name attribute.

### TECHNICAL STEPS
Inverted values of temperature grid.

Fixed _parse_val method. Previously 3 or 4 length string was considered a vector and causes error in case a string of the length was passed as an attribute in Principled Volume Shader.
Added support for 'velocity' and 'heat' attributes.
Improved get_domain_resolution function so 'velocity', 'heat' don't depend on noise_scale parameter.
Added condition if an object is vdb or domain. Now if the density grid is zeros volume material is empty instead of volume cube.
Changed condition for emission_grid_node.


VascoPi:
> It's interesting the value is inverted. Do you know why?

I agree this is strange, API Doc tells that `Smoke temperature grid, range 0 to 1 represents 0 to 1000K`, but while debugging found that values being retrieved from a domain are higher at smoke and lower at the fire.

bnagirniak (approved):
Looks ok

VascoPi:
@ARozghon could you please take a look at the last changes?
Added condition if an object is vdb or domain. Now if the density grid is zeros volume material is empty instead of volume cube.
Changed condition for emission_grid_node.

ARozghon (approved):
approved by QA

DagerD (approved):

bnagirniak (approved):

PR #519, opened 2022-06-07
RPRBLND-2148: Smoke simulation issue with Domain scale.

VascoPi:
### PURPOSE
Smoke simulation issue with Domain scale.

### EFFECT OF CHANGE
Fixed domain scale issue.

### TECHNICAL STEPS
Applied dynamic ratio obj.dimensions / obj.scale instead of constant.


bnagirniak (approved):
Looks good

ARozghon (approved):
looks good with the latest update

src/rprblender/export/volume.py
```diff
@@ -32,10 +32,10 @@ def key(obj: bpy.types.Object, smoke_modifier):
 
 
 def get_transform(obj: bpy.types.Object):
-    # Set volume transform. Note: it should be scaled by 2.0
     transform = object.get_transform(obj)
     scale = np.identity(4, dtype=np.float32)
-    scale[0, 0], scale[1, 1], scale[2, 2] = 2.0, 2.0, 2.0
+    scale[0, 0], scale[1, 1], scale[2, 2] = (obj.dimensions[i] / obj.scale[i] if obj.scale[i] > 0 else 0.0
```
bsavery, line 37: Should this have an absolute val?  What if obj.scale is -1

VascoPi: You're right, fixed that. Now while the negative scale values behavior is the same as Cycles.

PR #526, opened 2022-06-21
RPRBLND-2147: cryptomatte AOV picker improvement

VascoPi:
### PURPOSE
Cryptomatte AOV picker improvement. Add tooltip with object/material name for eyedropper tool.

### EFFECT OF CHANGE
Improved eyedropper tool with tooltip, it shows the name of the current element.

### TECHNICAL STEPS
Added methods sync_cryptomatte_hash, sync_obj_hash,_sync_mat_hash to context.py, for generating manifest data.
Added methods stamp_data_add_field and add_cryptomatte_metadata which allows adding metadata to render the result.
Added murmurh3_32 to utils/conversion.py, currently, it's the only hash function used within Cryptomatte.

### NOTES FOR REVIEWERS
Currently, the name for material to set name (rprObjectSetName ) is taken from node_parser.key() which is unique for every material even if it is a duplicate. It causes wrong behavior of cryptomatte material pass. The same materials have different hashes as a result these materials are different within cryptomatte material pass but should be the same. 
The same thing with instance.key() all instances keys are unique. It causes wrong behavior of cryptomatte object pass for instances. The purpose of the unique object name is LOOKUP_PRIMITIVE_RANDOM_COLOR, the random seed is generated from the shape name ( rprObjectSetName ).

Example of display name for materials -> "Material_Cube.004_Surface_Principled BSDF_BSDF"
Example of display name for instances  -> "Cube.001_603100188"
Example of display name for objects -> "Cube"

It would be good to have [another task](https://amdrender.atlassian.net/browse/RPRBLND-2167) on how to fix cryptomatte material pass and instances as well.


bnagirniak (approved):
Tested by QA - works good.

src/rprblender/engine/context.py
```diff
@@ -220,6 +224,28 @@ def resize(self, width, height):
...
+
+        if view_layer.rpr.crytomatte_aov_object:
+            self._sync_obj_hash()
+
+    def _sync_obj_hash(self):
+        for key, value in self.objects.items():
+            if not isinstance(value, (pyrpr.Mesh, pyrpr.Instance)):
+                continue
+
+            obj_name = str(key).replace("(", "").replace(")", "").replace(", ", "_").replace("'", "")
```
DagerD, line 241: Perhaps this should be @property or just a function?

ARozghon (approved):
working fine

src/rprblender/engine/context.py
VascoPi: Done.

DagerD (approved):

bsavery:
Why can't we use the native blender tools?

VascoPi:
> Why can't we use the native blender tools?
Could you please clarify what native blender tools?

PR #537, opened 2022-07-21
RPRBLND-2167: Cryptomatte material passes incorrect behavior, same material has different hash/color

VascoPi:
### PURPOSE
Cryptomatte material passes incorrect behavior, same material has different hash/color.

### EFFECT OF CHANGE
Improved eyedropper name hint for material cryptomatte passing material name.

### TECHNICAL STEPS
Fixed material cryptomatte identifier via passing material name to rprObjectSetName() instead of material.key().
Improved eyedropper name hint for material cryptomatte passing material name.


ARozghon (approved):
approved by QA

bsavery (approved):

PR #545, opened 2022-10-06
RPRBLND-2206: Check optimize Blender 3.3 splash screen

VascoPi:
### PURPOSE
Check optimize Blender 3.3 splash screen and various errors with HybridPro.

### EFFECT OF CHANGE
Fixed various bugs for Interactive renderer and Blender 3.3 splash.
Speed up sync process for hair particles of types 'None', 'Object', and 'Collection'.

### TECHNICAL STEPS
Remove unnecessary hair sync if render type is other than PATH. In case if render type is OBJECT or COLLECTION it syncs as an instance.
Remove support of Bump Node for HybridPro as it's not supported yet on the engine side.
Handle with nodes bl_idname == "NodeUndefined" (deprecated, outdated).

### NOTES FOR REVIEWERS
It should work with SDK 2.2.15.

To render the scene with HybridPro some of objects need to be excluded (for instance one of the islands)
otherwise got an error -18 after sync is finished.
Here’s a [ticket](https://amdrender.atlassian.net/browse/RPRHYB-719) with a trace.


DagerD (approved):
Tested. Works ok.

bsavery (approved):

ARozghon (approved):

PR #547, opened 2022-10-06
RPRBLND-2208: HybridPro add support for OCIO

VascoPi:
### PURPOSE
HybridPro add support for OCIO.

### EFFECT OF CHANGE
Added support of OCIO for Interactive render.

### TECHNICAL STEPS
Changes made to code.

### NOTES FOR REVIEWERS
Should work SDK 2.2.16 and higher.


DagerD (approved):

bnagirniak (approved):
Looks good

src/rprblender/export/image.py
```diff
@@ -122,7 +123,7 @@ def sync(rpr_context, image: bpy.types.Image, use_color_space=None, frame_number
 
 def set_image_gamma(rpr_image, image, color_space, rpr_context):
    # RPRImageTexture node color space names are in caps, unlike in Blender
-    if isinstance(rpr_context, context.RPRContext2):
+    if isinstance(rpr_context, (context.RPRContext2, RPRContextHybridPro)):
```
bsavery, line 126: Is this check necessary anymore?  I thought we removed old modes?

VascoPi: This check is made for High, Medium, and Low.

bnagirniak: We still support Hybrid only if HybridPro isn't supported
https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/blob/develop/src/rprblender/engine/__init__.py#L103
Therefore such check is required.

PR #557, opened 2022-11-02
RPRBLND-2221: RAM leak with multiple material set with Displacement only

VascoPi:
### PURPOSE
RAM leak with multiple material set with Displacement only.

### EFFECT OF CHANGE
Fixed various Blender hangs.

### TECHNICAL STEPS
Moved the parameter assignment to export_subdivision now it works with the final render. UI params have higher priority and are used if enabled.
Change order sync_visibility and assign_materials as in sync method.
Set level = 5 for displacement.

### NOTES FOR REVIEWERS
Implicit subdivision assignment didn't work for displacement at all It was always overridden to None after the assignment, the only case it starts working is when material update while viewport render.
This PR turns on subdivision when displacement is used.
Would be great to test various displacement scenes previously subdivision wasn't applied to the mesh and now results may vary.


bsavery (approved):

ARozghon:
not approved by QA. now with this fix, we have improved the result with one scene and created 5 new issues with other scenes.

ARozghon (commented):
Race spaceship is fixed but that provides an issue with other scenes

PR #583, opened 2023-02-22
RPRBLND-2247: Vector Math Hidden sockets issue

VascoPi:
### PURPOSE
Unlinked node affects render result.
Some of the nodes have few outputs and hide them according to condition. If the output was linked and then hidden the link still exists and affects the material.
image

### EFFECT OF CHANGE
Fixed unlinked node doesn't affect the render result.

### TECHNICAL STEPS
Added verification of whether a link is hidden.


bnagirniak (approved):
Looks good

DagerD (approved):

PR #586, opened 2023-02-27
Core 3.01.0 and RPRBLND-2251: Enable HIP / CUDA with RPR 3.0

VascoPi:
### PURPOSE
Update RadeonProRenderSDK to core_3.01.00.
Enable HIP / CUDA with RPR 3.0.

### EFFECT OF CHANGE
Switched Final render to precompiled kernels HIP/Cuda, from the user perspective now the first render after installation starts immediately.

### TECHNICAL STEPS
Updated RadeonProRenderSDK to the latest master Core 3.01.0.
Added `hipbin` dir, it contains precompiled HIP/CUDA kernels.
Adjusted create_zip_addon.py.
Fixed build process excluding `rprSetLogFunction`, `RPR_GET_SUPPORTED_DEVICES_FUNC_NAME`.
Added `hipbin_dir` function, returns HIP bin folder path.
Improved normal map export for HybridPro.

Improved condition for unsupported GPUs for HybridPro
Vulkan is not supported by GPU
`pyrpr.ERROR_INTERNAL_ERROR == Error: Fatal error: Unable to create Vulkan device, VkResult=-8`

for instance 2gb GPUs
`pyrpr.ERROR_OUT_OF_VIDEO_MEMORY == Error: Can't create pool, VkResult=-2`

Disabled RESTIR GI because it causes incorrect render results. Only regular RESTIR is enabled.
`RPR_CONTEXT_RESTIR_GI, RPR_CONTEXT_RESTIR_GI_BIAS_CORRECTION`

Improved warning.

### NOTES FOR REVIEWERS
There's no option for users to switch back to OpenCL in the PR let me know if we need it here.

`RPR_GET_SUPPORTED_DEVICES_FUNC_NAME` is excluded and not available due to an issue while building the addon.
error C2296: '|': illegal, left operand has type 'const char [23]'

Increased size of addon due to `hipbin`:
compressed - 326mb (was 106mb)
decompressed - 823mb (was 225mb)


bsavery:
We will need the option to use OpenCL eventually, for purposes of testing, probably not needed. 

I'm not sure we need a setting for the HIP kernel directory in the UI.  Why would the user ever change this?

VascoPi:
From the user's perspective, there's no need to change the path.
If we don't need it for purposes of testing, I'll remove it.

VascoPi:
Improved normal map export for HybridPro.

bnagirniak (approved):
Looks good.
@bsavery one question is about big size of hipbin folder, it is 600MB. It increases zip file from 100MB to 300MB. Do we plan to include it into zip addon?

bsavery:
I don't see any way to not include it, do you?

DagerD (approved):

bnagirniak:
> I don't see any way to not include it, do you?

I think we addon can download necessary binaries depending of user's card after addon activation. But we can think about it after release.

bsavery (approved):

ARozghon (approved):
Approved by QA

Issue #576, opened 2023-02-01
Improvements to Cuves / hair required


lenhone:
## Issue

#549 added rudimentary support for the new "curves", however so few features are supported it is not possible to get similar results to eevee or cycles. 

## Steps to Reproduce

Switch between eevee and RPR rendering.
[new hair blend.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/files/10558715/new.hair.blend.zip)

## System Setup (Environment)

OS: Linux

GPU: RX6800

Blender: 3.4.1, Plugin 3.5.2

## Detailed Description

Please consider implementing support for more features of the new hair system.

- "Strip" Curve rendering (as per render properties -> Curves -> Shape)
- "Subdivision"  (like render properties -> Curves -> Additional Subdivision or via geometry nodes)
- "Set Curve Radius" node for geometry nodes.
- "Curves Info" node for the material shader
- "Attribute" node for the material shader

Below is a comparison of RPR (top) vs eevee (bottom) to show the disparity of the limited support.

![eevee vs rpr](https://user-images.githubusercontent.com/96949377/216087508-d34895d6-7c7d-4dfd-9075-0dfd6c60de2e.png)

DagerD:
Hi.

Thank you. We will do investigation on this issue.

ARozghon:
Hi, this issue was fixed, please check with the latest weekly or release build and close this issue. Thank you

lenhone:
Thanks for letting me know an update has been made. 

Please note the following was tested with
Blender: 3.5.1
RPRBlenderPlugin: 3.6.2

Please see my check of the original issue items in the table below. Some have been resolved, some have not. If the following items are to be resolved with the upcoming [Hydra render engine, RPR delegate](https://projects.blender.org/blender/blender/pulls/104712) please close this log as all I'm after is support for the various node and shader options provided in blender).

| Request |Current Status  | Image |
| --- | ------------- | ------------- |
|"Strip" Curve rendering (as per render properties -> Curves -> Shape) | It appears this is not implemented | ![image](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/assets/96949377/e92f35fe-ff72-424c-a128-0631c366bf16)  |
| "Set Curve Radius" node for geometry nodes. | Works, eevee strip is same width as RPR |   |
[...]

PR #587, opened 2023-03-02
RPRBLND-2240: Update to complete support for curves

VascoPi:
### PURPOSE
User request https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/issues/576
Update to complete support for curves.
Add radius and UV support.

### EFFECT OF CHANGE
Now user can set the radius using the node "Set Curve Radius".
UVs now supported via the automatically generated attribute `surface_uv_coordinate`.

**Old behavior**
blender_ksKdUugyJ1

**New behavior**
blender_Dq8nP9VA1i

### TECHNICAL STEPS
Added support for `radius` attribute.
Added `iter_all_segments_radii` function which does the same thing as `iter_segments_radii`, but for all curves. This is because the radius may vary for different curves and control points.
Added support for `surface_uv_coordinate` to obtain UV coordinates.
Added comments.
Code adjustments.

### NOTES FOR REVIEWERS
Gaps between segments are also observed in other plugins. This will be discussed with the core team.
The ability to create a gradient along the curve will be added with the [ticket](https://amdrender.atlassian.net/browse/RPRBLND-2248)


DagerD (approved):

bnagirniak (approved):
Looks good. Waiting for QA to approve.

ARozghon (approved):
approved by QA

PR #592, opened 2023-03-13
Instance meshes where the "Data" mesh is reused between objects

bsavery:
### PURPOSE
Currently Objects which share a data mesh, we make a new mesh each time, this could be slow to export.

### EFFECT OF CHANGE
With each mesh we check if it's unmodified and if the mesh is already exported to reuse and instance.

### TECHNICAL STEPS
To do this we keep a dictionary of mesh masters

### NOTES FOR REVIEWERS
TODO does not yet handle viewport updates.


src/rprblender/export/mesh.py
```diff
@@ -350,55 +350,73 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, **kwargs):
     mesh = kwargs.get("mesh", obj.data)
     material_override = kwargs.get("material_override", None)
     smoke_modifier = volume.get_smoke_modifier(obj)
-
+
```
VascoPi, line 353: extra spaces

src/rprblender/export/mesh.py
```diff
@@ -350,55 +350,73 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, **kwargs):
     mesh = kwargs.get("mesh", obj.data)
     material_override = kwargs.get("material_override", None)
     smoke_modifier = volume.get_smoke_modifier(obj)
-
+    
     indirect_only = kwargs.get("indirect_only", False)
     log("sync", mesh, obj, "IndirectOnly" if indirect_only else "")
-
+
```
VascoPi, line 356: extra spaces

VascoPi (approved):
Works as intended, just minor fixes.
Regarding the viewport update, any mesh edits in edit mode don't affect the render result.

VascoPi:
The issue needs to be investigated, Q&A observed NorthStar crash of one of the test scenes.
Narrowing down the issue:
- The issue occurs when the emission shader is assigned to the curve object type.
- It is reproduced using Blender >= 3.0.

bsavery:
I guess I'm not clear where that code path would get used.  Can you take a look @VascoPi

VascoPi:
Fixed the crash and improved the instance's transform update.
Waiting for autotest results.

bnagirniak (approved):
Looks good, waiting for QA for test

ARozghon (approved):
Approved by QA

PR #594, opened 2023-03-20
RPRBLND-2265: viewport render got an error if borders is out of bounds

VascoPi:
### PURPOSE
Viewport render got an error if borders are out of bounds. In some cases render resolution width or height is zero. 

### EFFECT OF CHANGE
Fixed error if the render border is located out of the screen.

### TECHNICAL STEPS
Now if the width or height is zero we don't render and draw the image.


ARozghon:
all working fine, except for one case: if active upscale in Hybrid Pro that's provides weird GPU activity with the throttling and system short freezes 
image

VascoPi:
In case if render border is outside the screen the render resolution is like width = 33 and height = 0. It causes "division by zero" in other parts of code the decision. Changes that are made affect the minimum size of the render result >=1.
Memory leak related to another issue.
https://amdrender.atlassian.net/browse/RPRBLND-2267

src/rprblender/engine/viewport_engine.py
```diff
@@ -113,7 +113,7 @@ def __init__(self, context: bpy.types.Context):
 
         # getting render resolution and render border
         width, height = x2 - x1, y2 - y1
-        self.border = (x1, y1), (width, height)
+        self.border = (x1, y1), (max(width, 1), max(height, 1))
```
bnagirniak, line 116: Try to add check of width/height == 0 in draw() function

DagerD (approved):

bnagirniak (approved):

ARozghon (approved):
approved by QA. need to create a separate ticket for the FSR issue

PR #599, opened 2023-04-05
RPRBLND-2268: emission on Grease pencil provides blender crash

VascoPi:
### PURPOSE
In general, the issue is related to the new approach of instances sync (https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/pull/592)
The issue is that we create `pyrpr.Instance` from `pyrpr.Instance`. It's not correct and causes the errors below:
- Emission on Grease pencil provides blender crash.
- AttributeError: 'Instance' object has no attribute 'poly_count'

### EFFECT OF CHANGE
Fixed crash if extruded curve object applied with emission material.
Fixed error `'poly_count'` after the final render ends.

### TECHNICAL STEPS
Made a condition for `obj` variable in instance.py. now it's `instance.object` or `instance.instance_object`.
Fixed argument position in `instance.sync_update`.
Code improvements.


bnagirniak (approved):
Looks good

PR #601, opened 2023-04-10
RPRBLND-2275: instancing issue in RPR

VascoPi:
### PURPOSE
If we link mesh data from another blender file we can observe that the blender scene contains two meshes with the same name. To distinguish two meshes in `mesh_masters` we use `name_full` as the mesh key.
image

### EFFECT OF CHANGE
Fixed incorrect render results for scenes with linked data.

### TECHNICAL STEPS
Use `name_full` for the mesh key for `mesh_masters`.


DagerD (approved):

bnagirniak (approved):

ARozghon (approved):
Approved by QA

PR #612, opened 2023-04-21
Correct improper range in ShaderNodeHueSaturation

jombo23:
### PURPOSE
The Hue value being used is mapped 0-1 in blender, and 0-2pi in RPR, and is also reversed. This small edit fixes that to align with the blender norms

### EFFECT OF CHANGE
Anything that uses the HSV shader node should generate matching colors in ProRender and eevee/cycles. Hopefully this is a quick and painless correction :)

### TECHNICAL STEPS
Since RPR is calling sin/cos directly on the hue value, its expected that the inputs are 0-2pi. Blenders norm is 0-1, and are also inverted compared to RPR. Therefore, the incoming hue value is now multiplied by negative tau in order to achieve matching color generation with the rest of the rendering engines.

### NOTES FOR REVIEWERS
Sample blend file for comparison
[prorenderScaleDemo.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/files/11293626/prorenderScaleDemo.zip)

Renderer outputs for comparison:

Eevee
EEVEE Hue

Cycles
Cycles Hue

ProRender
prorender hue

###### ProRender ***After*** this patch ######
prorender patched hue

Please note that the colors do not seem to be identical, but as far as my testing goes, they are. Prorender just represents colors differently (unless theres some other math bug, but i do not believe that is the case).

### DIFF
--- a/src/rprblender/nodes/blender_nodes.py
+++ b/src/rprblender/nodes/blender_nodes.py
@@ -2711,7 +2711,7 @@ class ShaderNodeHueSaturation(NodeParser):
 
         color = self.get_input_value('Color')
         fac = self.get_input_value('Fac')
-        hue = self.get_input_value('Hue') - 0.5
+        hue = (self.get_input_value('Hue') - 0.5) * -math.tau
         saturation = self.get_input_value('Saturation')
         value = self.get_input_value('Value')
 


bsavery (approved):

VascoPi (approved):
Verified.
The remap 0-2pi looks correct according to [16-hsv-color-transform](http://beesbuzz.biz/code/16-hsv-color-transforms) reference.
Waiting for approval from the Q&A.

VascoPi:
It's ready for tests.
@jombo23 could you please update with latest master?

jombo23:
@VascoPi Done.

ARozghon (approved):
Approved by QA

Issue #615, opened 2023-04-23
RPR ignores Material Output target


jombo23:
Every issue should include 3 things:
1.  System setup (GPU, CPU, OS version, Blender version)
     Win11, Vega64, xeon e5 2696v4, blender 3.5.0

Example blend file
[prorendermats.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/files/11302565/prorendermats.zip)

In the material editor, RPR uses the material output that was last clicked. This makes having preview materials for eevee impossible, as different things cause the material outputs to change order, and youll randomly get an eevee material wrapped up in your prorender ones.

Ultimately, adding an RPR target to the material output node would likely be the best way around this, but that would exclude it being added in projects when prorender wasnt set as the renderer, however, at the very least, any target besides ALL should be ignored for the sake of the materials system working as intended.

![68e505d6-cb78-4d70-be89-cda78a59eef06a0c9bbe56](https://user-images.githubusercontent.com/4957057/233812867-b70c36ad-858e-4909-9a62-52d0c28a85ef.gif)

PR #616, opened 2023-04-23
Reject other renderer Material Outputs (#615)

jombo23:
### PURPOSE
This PR makes RPR reject material output nodes from other renderers. (see #615 )

### EFFECT OF CHANGE
RPR will no longer grab ShaderNodeOutputMaterial nodes with targets set to other rendering engines. This enables users to have materials with different sets of node paths for different renderers.

### TECHNICAL STEPS
The conditions for locating a material output node were changed from;
node.bl_idname == 'ShaderNodeOutputMaterial' and node.is_active_output

to
node.bl_idname == 'ShaderNodeOutputMaterial' and node.target == 'ALL'

so that material outputs for other renderers are ignored.

### NOTES FOR REVIEWERS
I am not sure why "is_active_output" was originally on here, but from my testing, it seems irrelevant, and messes up the condition of the target renderer, as if nodes get reordered from user interaction, a node can be "inactive" but still be completely valid, and the proper path.

In my testing I have yet to come across a scenario where this breaks anything.


src/rprblender/export/material.py
```diff
@@ -36,7 +36,7 @@ def get_material_output_node(material):
         return None
 
     return next((node for node in material.node_tree.nodes
-                 if node.bl_idname == 'ShaderNodeOutputMaterial' and node.is_active_output),
+                 if node.bl_idname == 'ShaderNodeOutputMaterial' and node.target == 'ALL'),
```
bsavery, line 39: Thinking about this, we should also accept Cycles target?

jombo23: In an instance where both exist, prioritizing all should always be first, unless there's a way to add prorender to the list of rendertargets without redefining the output class (so it doesn't ONLY show prorender in an output node placed while the renderer is set as prorender).

In any other case though really, it's dangerous to go with a renderer specific to any other renderer. If someone specifically specifies cycles, why should prorender use it?

bsavery: There is not a way to add ProRender to the list of outputs.

Generally speaking we try to translate eevee / cycles nodegraphs into something RPR can use.  The case that we're really talking about here if:
1.  The user does not have an "All" output node.
2.  The user has two output nodes, one for Cycles and one for EEVEE.  

In that case which should we choose to use?  There's no obvious correct answer.  So I'm ok with just the solution you have here.  

Of course you could do something like:
 Look for the first 'ALL" output node, if none found choose the first "Cycles" output, if not then whichever EEVEE.  
 
 Willing to hear opinions on this though.

jombo23: aha! I get what you mean now.

Perhaps when a non preferred output is taken it can warn a user in the log?

Just thinking about the case where I (and a handful of other people I know) run two sets of nodes per material for previews and renders, it would be nice to know if the preferred output was not available.

PR #618, opened 2023-04-25
Add Vertex attribute support

jombo23:
Here goes nothing...

### PURPOSE
Adding vertex attribute support to RPR 

### EFFECT OF CHANGE
Named Vertex attributes such as colors, vectors, and floats will now be usable in materials.

### TECHNICAL STEPS
Added mesh_attribute_names to RPRContext
Added set_primvar to pyrpr
Added a basic ShaderNodeAttribute Implementation that uses prior mesh_attribute_names
Added ShaderNodeAttribute into the input nodes list in rpr_nodes
Changed ShaderNodeLookup "Vertex Colors" to look up the new primvar version
Added basic attribute parsing to the mesh export

### NOTES FOR REVIEWERS
There are some comments where I couldnt reasonably figure things out.
Bools, strings, and ints are not passed through as primvars, as I wasnt sure what the proper way to handle them in the material node section would be regarding rounding the numbers/clamping/rolling/etc, and of course, where could strings possibly go.

ShaderNodeAttribute doesnt care what 'attribute_type' is currently, and just uses the value straight from prorender.

The attributes are stored PER VERTEX, I did not implement handling them per face, as that requires a triangulated mesh, and I wasnt sure how to do that and maintain face order with the rest of the renderer. 

There might be redundant casts in set_primvar in pyrpr. I figured cast > no cast. Python isnt my language

Updating attributes AFTER enabling viewport rendering causes them to all read 0. Im not sure how to fix the desync. This doesnt impact renders as I guess its synced every frame.

Enabling viewport rendering AFTER switching to edit/paint mode causes an out of bounds access in export.mesh.vertex_attributes because the attributes still exist but the data doesnt when youre in edit/paint mode? 

I probably did things totally wrong with horrible syntax, but I honestly dont know any better, this is just about the second time I've ever touched python, so feel free to let loose with the criticism, I dont think what I did was done particularly well.

**This is in multiple commits because I did this through the web interface instead of updating my local prorender build, so its likely that I may have missed an indent somewhere, or forgotten some line of code, but I dont \*\*think\*\* I did.** 

This is mostly for review only, and ill likely set up a proper fork locally if needed, So I'm marking this as a draft. 

acb6d746-5075-4fe2-a835-a4b496d96e2f70d04a076d

tada


src/bindings/pyrpr/src/pyrpr.py
```diff
@@ -561,14 +561,23 @@ def mark_static(self, is_static):
     def set_vertex_value(self, index: int, indices, values):
         ShapeSetVertexValue(self, index, ffi.cast("rpr_int *", indices.ctypes.data),
                             ffi.cast("float *", values.ctypes.data), len(indices))
-
+    
+    def set_primvar(self, key: int, data, floatcount: int, count: int, interop):
```
bsavery, line 565: rename some variables: 
float_count
interop -> interp or interp_type

src/bindings/pyrpr/src/pyrpr.py
```diff
@@ -561,14 +561,23 @@ def mark_static(self, is_static):
...
-        indices = np.arange(len(colors), dtype=np.int32)
-
-        # index is 0-3 index (use for r,g,b,a)
-        for i in range(4):
-            values = np.ascontiguousarray(colors[:, i], dtype=np.float32)
-            self.set_vertex_value(i, indices, values)
+        vColors = np.array(colors)
+        vColors = vColors.flatten()
+        vColors = np.ascontiguousarray(vColors, dtype=np.float32)
+        self.set_primvar(99,vColors,  len(vColors), 4, PRIMVAR_INTERPOLATION_VERTEX)
```
bsavery, line 580: Why is the key here "99" ? I'm assuming this is a test.  You could make a dictionary of primvar names -> number

src/rprblender/export/mesh.py
```diff
@@ -112,6 +113,38 @@ def init_from_mesh(mesh: bpy.types.Mesh, calc_area=False, obj=None):
 
         if calc_area:
             data.area = sum(tri.area for tri in mesh.loop_triangles)
+           
+        # Not possible to store strings/ bools in primvars. If the type was bool, it would be possible to just pass 1 or 0, and let it be interpolated, and then round it, but that would require extra code in the shaderattribute node
+        # Also cannot currently handle anything besides vertex attributes (point/corner), without triangulating the mesh first, which is over my head.
+        # Ints and Int8s are currently excluded as well, and byte color is stored as float anyway(?)
+        # Ignoring givens: uvmap and position
+        # Returning list of lists, [AttName, AttSize, AttData, AttInterop] Interop stored for future implementation Face/Edge types (im assuming which are far less common)
+        validAtts =  [e for e in mesh.attributes if e.data_type in ['FLOAT', 'FLOAT_VECTOR', 'FLOAT_COLOR', 'BYTE_COLOR', 'FLOAT2']
```
bsavery, line 122: again use snake case valid_attrs for variable names

src/rprblender/export/mesh.py
```diff
@@ -112,6 +113,38 @@ def init_from_mesh(mesh: bpy.types.Mesh, calc_area=False, obj=None):
 
         if calc_area:
             data.area = sum(tri.area for tri in mesh.loop_triangles)
+           
+        # Not possible to store strings/ bools in primvars. If the type was bool, it would be possible to just pass 1 or 0, and let it be interpolated, and then round it, but that would require extra code in the shaderattribute node
+        # Also cannot currently handle anything besides vertex attributes (point/corner), without triangulating the mesh first, which is over my head.
```
bsavery, line 118: I guess I'm not clear here.  Why would you need to triangulate a mesh?  Blender gives you attribues in what interpolation?  Vertex / Uniform / Constant?  

Anything thats varying across vertices is handled at the renderer to interpolate.

src/rprblender/export/mesh.py
```diff
@@ -405,6 +438,12 @@ def sync(rpr_context: RPRContext, obj: bpy.types.Object, **kwargs):
 
         if data.vertex_colors is not None:
             rpr_shape.set_vertex_colors(data.vertex_colors)
+            
+        if data.vertex_attributes is not None:
+            for att in data.vertex_attributes:
+                if att[0] not in rpr_context.mesh_attribute_names:
+                    rpr_context.mesh_attribute_names.append(att[0])
```
bsavery, line 445: Personally I would do it such that the mesh_attribute_names is a dictionary where the key:value is 

attribute_name : index

src/bindings/pyrpr/src/pyrpr.py
jombo23: will do, was mostly following the names from the sdk. I thought interop was weird as well but didn't want to use and other term that was used somewhere else (which is why I figured they did it)

src/bindings/pyrpr/src/pyrpr.py
jombo23: I used 99 here based off of the sdk again. In terms of vertex color attributes, the name can be different between two meshes. 

eg, mesh1 vertex color's attribute might be named "Attribute" (default) and mesh2's might be named "color" 
there could even be a mesh3 which has two color attributes. one named "Attribute" and one named "color".

Assigning a static value to the active color attribute is the only way I can see to keep the active stuff active as the 'vertex color' attribute referenced elsewhere.

figure mesh1 would get 100, mesh 2 would get 101, and mesh 3 would have 100 and 102. How would the vertex color attribute know which to pick unless another list of what the active color attributes are per mesh was made.

I could be entirely misunderstanding, but that's how I interpreted it.

The sdk example started at 100, so so did I.

src/rprblender/export/mesh.py
jombo23: This was a slip by me, thanks. I normally write in camelcase with underscores for prefixing the "class" the function would belong to, lol.

src/rprblender/export/mesh.py
jombo23: In blender if you have a cube, and create an attribute with the domain of "face" you will get six values across the entire cube. The mesh is triangulated before it is sent to prorender, and I'm not sure how to get triangle level attributes out of blender *after* triangulation occurs. if you try to submit the six values to prorender as primvars, it throws a parameter error because the total values don't match the mesh triangle count

src/rprblender/export/mesh.py
jombo23: I'm not entirely sure of the proper way to go about this.

VascoPi: Probably it's a good idea to use `key: value`, where `key` is the combination of a mesh key and attribute name, and the `value` is an index.
To make it easy to access attributes of a specific mesh during sync_update is the purpose.

VascoPi:
Added the updates in https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/discussions/617#discussioncomment-5771404

PR #619, opened 2023-04-25
RPRBLND-2280: issue with instances with Blender 2.93LTS scene.

VascoPi:
### PURPOSE
In case a scene contains data items with the same name, while synchronization a rpr_shape in mesh_masters  can be overridden with another one.
For instance, we have mesh and curve objects with the same names. In both cases, obj.data.name_full is the same but the types and data are different.
 

### EFFECT OF CHANGE
Fixed issue with Blender 2.93LTS scene. (regression https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/pull/592)

### TECHNICAL STEPS
Implemented `key` function for the mesh module using `obj.original` to get the initial object type and combine the key string.
Adjusted the code.


DagerD (approved):

bnagirniak (approved):
Looks good

Issue #626, opened 2023-05-26
Crash upon exiting Blender


anarchymedes:
Issue: 
Upon closing Blender, an error message pops up, stating that the application 'quit unexpectedly' (the usual MacOS crash routine).

Steps to reproduce:
- Download and instal Radeon Pro v.3.6
- Enable the add-on in the Blender Settings.
- No need to open any file: the scene may remain the default one; also, no need to change the render engine to Radeon Pro Render: the result is the same; quit Blender – or try to disable the Radeon Pro Render add-on. Observe the crash.

System: 
Mac Mini 2023, M2 Pro, MacOS Ventura 13.3.1(a), Blender 3.5.1 (BTW, this issue has been present in the previous versions of both Radeon Pro Render and Blender (starting from 3.4, I think), as well as in the Intel-based version of RPR: I was hoping that this release would finally fix it)

The crash dump (the part the length limit has allowed me to include) follows:
Process:               Blender [4617]
Path:                  /Applications/Blender.app/Contents/MacOS/Blender
Identifier:            org.blenderfoundation.blender
Version:               3.5.1 (3.5.1 2023-04-25)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2023-05-26 18:32:45.8744 +1000
OS Version:            macOS 13.3.1 (22E772610a)
Report Version:        12
Anonymous UUID:        AA0EE291-3DB7-8C4E-1688-202DFCBCA29A

Time Awake Since Boot: 1200 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Codes:       0x0000000000000001, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [4617]

VM Region Info: 0 is not in any region.  Bytes before following region: 4334862336
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      10260c000-10d87c000    [178.4M] r-x/r-x SM=COW  ...MacOS/Blender

[...]

PR #631, opened 2023-06-09
RPRBLN-2281: Mac Mini M2 crash on addon deactivation

VascoPi:
### PURPOSE
Mac Mini M2 crash on addon deactivation.
https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderBlenderAddon/issues/626

### EFFECT OF CHANGE
Fixed addon deactivation crash on Mac Mini M2.

### TECHNICAL STEPS
Put apart un/registration for node, socket, and socket interface classes and changed its order.

### DIFF
--- a/src/rprblender/nodes/__init__.py
+++ b/src/rprblender/nodes/__init__.py
@@ -143,28 +143,34 @@ def hide_cycles_and_eevee_poll(method):
 from . import sockets
 from . import rpr_nodes
 
-register_classes, unregister_classes = bpy.utils.register_classes_factory([
-    sockets.RPRSocketColorInterface,
+register_socket_classes, unregister_socket_classes = bpy.utils.register_classes_factory([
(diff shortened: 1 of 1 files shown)


bsavery:
It's interesting that this would fix the problem, or rather that the different order only effects one platform.

bnagirniak (approved):
Looks good

VascoPi:
While debugging it was found that utilizing `bpy.utils.register_classes_factory`'s order for given classes causes the error.
Simply changing the unregistering order fixes the issue.

Eventually, I didn't find an accurate explanation for why it happens only at M2. Looks like the blender code doesn't have a special condition for M2 as well.

Also changing Python version to 3.10 while the building process doesn't give any results.

Agree it's strange behavior on M2 because no warnings or errors are observed on other systems using the common `bpy.utils.register_classes_factory`'s order for given classes.

PR #637, opened 2023-06-29
RPRBLND-2290: deactivate hided denoiser options

VascoPi:
### PURPOSE
Need to deactivate old denoiser options, because this option was hidden in UI, but is still active and the user can't switch off or set up.

### EFFECT OF CHANGE
Denoiser doesn't affect the final render result in old scenes.

### TECHNICAL STEPS
Permanently deactivate denoiser "enable" property.

### NOTES FOR REVIEWERS
[Ticket](https://amdrender.atlassian.net/browse/RPRBLND-2291) to remove redundant 'denoiser' code.


ARozghon (approved):
Approved by QA

bsavery (approved):

PR #674, opened 2024-03-26
Odilkhan Yakubov PR commits

odil24:
### EFFECT OF CHANGE
- Fix: removed params because of Viewport rendering;
- Fix: Unhide Upscale and Viewport denoising and minor changes;
- Improved: increased syncronization performance for Viewport/Material preview when mesh sync;
- Improved: viewport rendering.

PR #684, opened 2024-05-28
Feature/pr 674

takahiroharada:
From #674


feniksa:
Guys, 

Currently PR feels like mix of changes and it is hard to review. 

I think, the best will be to separately make "good" PR-674 and merge it.

PR #685, opened 2024-06-21
testing script

spenceau:
### PURPOSE
Test Script to automate testing for RPR Blender Plug-in

### EFFECT OF CHANGE
final_render.py to generate Final Render in background
compare_render.py to compare Final Render with Ground Truth using Pixel Wise MSE and SSIM
viewport_render.py to generate Viewport Render (using final as specified) via opening Blender in foreground
test_script.ps1 to automate all 3 steps for a Single Scene and Blender Version 

this is stored within cmd_tools/test_script

### TECHNICAL STEPS
first commit; see README

### NOTES FOR REVIEWERS
Notes, if needed.


spenceau:
needed to delete some old files

PR #686, opened 2024-06-21
RPR Test Script

spenceau:
### PURPOSE
Test Script to automate testing for RPR Blender Plug-in

### EFFECT OF CHANGE
cmd_render.py is a wrapper file that unzips the plugin and calls relevant files
add_script.py adds the subdir that extracted **rprblender** build to `bpy.context.preferences.filepaths.script_directories`
final_render.py to generate Final Render in background
remove_script.py to remove build subdir from `bpy.context.preferences.filepaths.script_directories` and copy over crash logs from LOCALAPPDATA in case of a failed render
compare_render.py to compare Final Render with Ground Truth using Pixel Wise MSE and SSIM
//viewport_render.py to generate Viewport Render (using final as specified) via opening Blender in foreground

run_render.cmd for Windows
run_render.sh for Ubuntu/Linux

this is stored within **cmd_tools/test_script**

### TECHNICAL STEPS
see README

### NOTES FOR REVIEWERS
cmd_tools/test_script/README.md specifies that this requires Python 3.11
viewport_render.py is deprecated as viewport render will probably not be used in the testing pipeline?


cmd_tools/test_script/viewport_render.py
```diff
@@ -0,0 +1,102 @@
...
+    except ImportError as e:
+        print(f"Error importing rprblender: {e}")
+    except Exception as e:
+        print(f"Error registering rprblender: {e}")
+
+
+def main():
+    print("Starting viewport render script...")
+
+    # import bpy
```
feniksa, line 61: I think, it is possible to drop this comment as argparse used below.

odil24:
@spenceau What PR does?

feniksa (approved):

build.cmd
```diff
@@ -64,14 +64,14 @@ if %vs_major%==15 or %vs_major%==16 (
...
-py -3.11 src\bindings\pyrpr\src\pyrprapi.py %castxml%
+python -m pip install cffi
+python -m pip install numpy
+python cmd_tools\create_sdk.py
+python src\bindings\pyrpr\src\pyrprapi.py %castxml%
 
 set bindingsOk=.\bindings-ok
 if exist %bindingsOk% (
-	py -3.11 build.py
+	python build.py
```
takahiroharada, line 74: Why? we shouldn't do these changes. Let's keep explicit python versioning.

takahiroharada (changes requested):
Keep the changes minimum. SDK update shouldn't be in this PR. If needed, we need to do in another PR. Please do review all the changes and remove unnecessary files from this PR.

spenceau:
Changed to merge fork into the RPRBLD-0-test-script

PR #713, opened 2025-07-27
[Node converter] From active material > Whole material convertion

odil24:
### PURPOSE
One-click solution

### EFFECT OF CHANGE
Old feature (button) only converted active material but with this patch it will convert all materials in the scene at once

### TECHNICAL STEPS
N/A

### NOTES FOR REVIEWERS
N/A


odil24:
@feniksa @takahiroharada Please, review

